PR #20, opened 2016-07-08
Checkstyle: fix build by removing modifiers

marnix:
Some very low hanging fruit to fix that poor Checkstyle build... :-)


smarr:
Thanks, for some reason, I used to have problems if the same ant process does the download and use the checkstyle plugin. Not sure why that was. Perhaps any ideas? Sorry, merge will take a bit of time.

marnix:
No, no idea why download-then-use would be different for one ant process or two.

smarr:
Thanks

PR #22, opened 2016-07-28
Initial support for actor-level debugging

smarr:
These changes include:

- moves actor tracing related features to ActorExecutionTrace
- adds BreakpointActor, which is used for the actor debugger
- eventual messages can now indicated that execution should pause before being processed
- add session package to debugger to contain classes related to a debugging session
- minor refactorings

/cc @ctrlpz


src/som/interpreter/actors/Actor.java
```diff
@@ -64,9 +66,15 @@ public static void traceActorsExceptMainOne(final SFarReference actorFarRef) {
   /** Is scheduled on the pool, and executes messages to this actor. */
   private final ExecAllMessages executor;
 
+  /**
+   * Contains the operations for the management of messages between actors.
+   */
+  private final LocalManager localManager;
```
smarr: This is something I don't want to integrate like that, because it makes creating actors more expensive. The local manager is a huge object. It has an object buffer, two, hash maps, and other state.

I think, we might want to put this into a subclass of actor that is only used when executing with a debugger.

Or, we change the whole approach in a way that the actors themselves do not have or do not need a local manager.

src/som/interpreter/actors/Actor.java
```diff
@@ -122,13 +130,21 @@ protected void logNoTaskForActor() { }
   @TruffleBoundary
   public final synchronized void send(final EventualMessage msg) {
     assert msg.getTarget() == this;
-    mailbox.append(msg);
-    logMessageAddedToMailbox(msg);
+    if (msg.hasPause()) {
+      VM.println("Message pause--");
```
smarr: Please do not commit debugging output.
In git it should be possible to select specific lines not to be included in a commit.

src/som/interpreter/actors/Actor.java
```diff
@@ -239,25 +239,15 @@ public ForkJoinWorkerThread newThread(final ForkJoinPool pool) {
   public static final class ActorProcessingThread extends ForkJoinWorkerThread {
     public EventualMessage currentMessage;
     protected Actor currentlyExecutingActor;
-    protected final ObjectBuffer<SFarReference> createdActors;
-    protected final ObjectBuffer<ObjectBuffer<EventualMessage>> processedMessages;
+    ObjectBuffer<SFarReference> createdActors;
+    ObjectBuffer<ObjectBuffer<EventualMessage>> processedMessages;
```
smarr: Why do these fields need to be mutable (i.e., why removing the `final`)?
And, why is `protected` not sufficient anymore?

src/som/interpreter/actors/EventualMessage.java
```diff
@@ -269,4 +272,14 @@ public static EventualMessage getCurrentExecutingMessage() {
...
     return ((ActorProcessingThread) t).currentMessage;
   }
+
+  //indicates if the message is already in the actor mailbox and it is paused (due to breakpoint or step into command)
+  public boolean hasPause() {
+    return pause;
+  }
+
+  //call when adding a breakpoint
+  public void setPause(final boolean value) {
```
smarr: Is this really needed, can't we make the field final and properly initialize the message on creation?

src/som/interpreter/actors/EventualSendNode.java
```diff
@@ -162,6 +162,8 @@ protected void sendDirectMessage(final Object[] args, final Actor owner,
       DirectMessage msg = new DirectMessage(
           EventualMessage.getCurrentExecutingMessage(), target, selector, args,
           owner, resolver, onReceive);
+
+      checkMessageBreakpointed(msg);
```
smarr: I think, this should be a function that takes the target actor, and returns a boolean for whether or not the message should trigger a breakpoint.

src/tools/debugger/FrontendConnector.java
```diff
@@ -209,36 +211,139 @@ public void sendActorHistory() {
...
+    }
     log("[ACTORS] Message sent waiting completed");
 
     sender.close();
   }
 
   public void requestBreakpoint(final boolean enabled, final URI sourceUri,
-      final int startLine, final int startColumn, final int charLength) {
+      final int startLine, final int startColumn, final int charLength,
+      final String role) {
```
smarr: Would be good to replace `role` with some enum or so. Passing strings around should generally avoided, I think. Also, having an enum would document what the proper options are.

src/tools/debugger/FrontendConnector.java
```diff
@@ -209,36 +211,139 @@ public void sendActorHistory() {
...
+       actor.getLocalManager().setFileName(sourceUri);
+       ActorExecutionTrace.assignBreakpoint(bp, actor, location, receiver);
+      }
+
     } catch (IOException e) {
       throw new RuntimeException(e);
     }
   }
 
+  private BreakpointId getBreakpointId(final Map<BreakpointId, Breakpoint> storeBreakpoints, final URI sourceUri,
```
smarr: This seems like functionality that belongs into the `Breakpoints` class, especially since you needed to add a getter to get all known breakpoints. I'd prefer to remove the getter, and have this id lookup being in `Breakpoints`.

src/tools/debugger/FrontendConnector.java
```diff
@@ -209,36 +211,139 @@ public void sendActorHistory() {
...
+        if (lb.equals(bId)) {
+          return lb;
+        }
+      }
+    }
+
+    return null;
+  }
+
+  private BreakpointLocation getBreakpointLocation(final URI sourceUri, final int startLine, final BreakpointId bId) {
```
smarr: This seems also functionality that does better belong into `Breakpoints`.

src/tools/debugger/FrontendConnector.java
```diff
@@ -266,6 +371,41 @@ public void shutdown() {
...
+  }
+
+/**
+ * Encapsulates the holder class and the method name of the breakpoint.
+ *
+ * @author carmentorres
+ *
+ */
+  public class BreakpointLocation{
+    private String holderClass;
```
smarr: This is problematic. The class name is not a unique identifier. Would be good to have another, more unique, criterium.

src/tools/debugger/WebDebugger.java
```diff
@@ -132,4 +134,8 @@ public void startServer(final Debugger dbg) {
     breakpoints = new Breakpoints(dbg);
     connector = new FrontendConnector(breakpoints, instrumenter, this);
   }
+
+  public Map<Source, Set<RootNode>> getRootNodes() {
```
smarr: I don't like all these getters. If you need only something specific, then add the query method here directly. It looks to me that you want to get the root nodes for a specific source, based on the source URI, and not all of them.

tools/controller.js
```diff
@@ -96,7 +96,7 @@ Controller.prototype.onToggleLineBreakpoint = function (line, clickedSpan) {
   this.view.updateLineBreakpoint(breakpoint);
 };
 
-Controller.prototype.onToggleMessageSendBreakpoint = function (e) {
+/*Controller.prototype.onToggleMessageSendBreakpoint = function (e) {
```
smarr: Why is this code comment out instead of being removed?
Unused code should be removed. If it is later needed again, we can find in git/on github

tools/view.js
```diff
@@ -395,8 +395,35 @@ View.prototype.displaySuspendEvent = function (data, getSource) {
     annotateArray(annotationArray, source.id, data.sections);
     sourceFile.html(arrayToString(annotationArray));
 
+    var sendOperator = sourceFile.find(".EventualMessageSend");
```
smarr: would be nice if this code could be put into its own function. we got here more than 10 lines of code. Think that justifies giving it a proper name and its own function :)

src/tools/debugger/FrontendConnector.java
```diff
@@ -266,6 +371,36 @@ public void shutdown() {
...
+   */
+  public class BreakpointLocation {
+    private String holderClass;
+    private String methodName;
+    private BreakpointId id;
+
+    BreakpointLocation(final String holderClass, final String methodName,
+        final BreakpointId id) {
+      this.holderClass = holderClass;
+      this.methodName = methodName;
```
smarr: Thinking about this, I don't think we want to identify the method by class and name.
Instead, we should just identify it by its source location. A `Method` has a source section. That one identifies it uniquely. And, also avoids having two classes with the same name and the same method name clash.

tools/view.js
```diff
@@ -440,9 +468,6 @@ View.prototype.displaySuspendEvent = function (data, getSource) {
     var annotationArray = sourceToArray(source.sourceText);
     annotateArray(annotationArray, source.id, data.sections);
     sourceFile.html(arrayToString(annotationArray));
-
-    // enable clicking on EventualSendNodes
-    enableEventualSendClicks(sourceFile);
```
smarr: Why is this one removed? I think this causes the problem with the breakpoints not working after a breakpoint was triggered

tools/controller.js
```diff
@@ -98,13 +98,22 @@ Controller.prototype.onToggleLineBreakpoint = function (line, clickedSpan) {
...
-      function (source) { return new SendBreakpoint(source, sourceSection); });
+  this.view.updateSendBreakpoint(breakpoint);
+};
+
+Controller.prototype.messageBreakpointSender = function (sectionId) {
+  dbgLog("--messageBreakpointSender");
+
+  var sourceSection = this.dbg.getSection(sectionId),
+  breakpoint    = this.toggleBreakpoint(sectionId,
+      function (source) { return new SendBreakpoint(source, sourceSection, "sender"); });
```
smarr: Hmm I don't like this code.
This is code duplication in those two methods, no?
The only difference is "receiver" vs "sender", right?

Would be good to have a method that takes that as a parameter.

Also, the method you replaced was name "onToggle", why is this naming not preserved for these methods?
This indicates that these methods are UI toggling event handlers.

tools/view.js
ctrlpz: I thought that the enableEventualSendClicks in showSource function is the one that triggers the popover on first load, so I think is not need it there anymore in displaySuspendEvent function I will test it

tools/controller.js
ctrlpz: I agree, I will fix it

tools/view.js
smarr: It is needed. Currently, when a suspended event is triggered, the whole display of the source file is essentially scraped and recalculated. This also requires reapplying the this part.

At some point, we should avoid doing that, but the idea is that some of the information we want to display is dynamic, so, it can change.

ctrlpz: I will save this change but the problem the connection refused error remains

src/som/interpreter/actors/EventualSendNode.java
```diff
@@ -256,5 +261,12 @@ protected boolean isTaggedWith(final Class<?> tag) {
       }
       return super.isTaggedWith(tag);
     }
+
+    protected boolean isMessageBreakpointed(final Actor receiver, final RootCallTarget root) {
+      if (receiver instanceof LocalManager) {
```
smarr: The test here should be on `VmSettings.TRUFFLE_DEBUGGER_ENABLED` instead.
The main reason is that the VmSettings field is a constant, which can be easily resolved by the compiler, so that the test does not have to be done at runtime in the non-debugging mode. This is very important for performance.

PR #26, opened 2016-08-08
Implement send-site receiver breakpoints

ctrlpz:
Todo tasks:

1. decide functions that are going to be use in BreakpointActor class
(send, schedule and serve for the receiver breakpoints?)

2. implement semantics for receiver breakpoints (I'm working on this)
 - [x] activate the breakpoint only once
 - [x] activate from an asynchronous message send
 - [ ] get location of the receiver breakpoint where the debugger has to stop

3. stepping for the receiver breakpoints

### DIFF
--- a/src/tools/actors/ActorExecutionTrace.java
+++ b/src/tools/actors/ActorExecutionTrace.java
@@ -75,22 +75,4 @@ public class ActorExecutionTrace {
     }
     return processedMessages;
   }
-
-  /**
-   * Assign breakpoint for the actor that is the receiver of the message.
-   */
-  /*public static void assignBreakpoint(final Breakpoint breakpoint, final Actor actor,
-      final BreakpointTrace breakpointTrace, final boolean receiver) {
-
-    ObjectBuffer<ObjectBuffer<SFarReference>> actorsPerThread = getAllCreateActors();
-    for (ObjectBuffer<SFarReference> perThread : actorsPerThread) {
-      for (SFarReference farReference : perThread) {
-        if (farReference.getActor().equals(actor)) {
-          Actor actorReference = farReference.getActor();
-          ((BreakpointActor) actorReference).addBreakpoint(breakpoint, breakpointTrace, receiver);
-          break;
-        }
-      }
-    }
-  }*/
 }
--- a/src/tools/actors/ActorExecutionTrace.java
+++ b/src/tools/actors/ActorExecutionTrace.java
@@ -1,15 +1,11 @@
 package tools.actors;
 
-import com.oracle.truffle.api.debug.Breakpoint;
-
 import som.VmSettings;
 import som.interpreter.actors.Actor;
 import som.interpreter.actors.EventualMessage;
 import som.interpreter.actors.SFarReference;
 import som.vm.ObjectSystem;
 import tools.ObjectBuffer;
-import tools.debugger.session.Breakpoints.BreakpointDataTrace;
-import tools.debugger.session.BreakpointActor;
 
 public class ActorExecutionTrace {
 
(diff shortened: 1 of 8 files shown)


src/tools/debugger/session/Breakpoints.java
```diff
@@ -195,6 +203,33 @@ public Breakpoint getAsyncMessageRcvBreakpoint(final URI sourceUri,
     return bp;
   }
 
+  public synchronized void saveReceiverBreakpoint(final URI sourceUri, final int startLine,
+      final int startColumn, final int charLength) {
+    SectionBreakpoint bId = new SectionBreakpoint(sourceUri, startLine, startColumn, charLength);
+    if (!receiverBreakpoints.keySet().contains(bId)) {
+      this.receiverBreakpoints.put(bId, null);
```
smarr: This can be simplified to `receiverBreakpoints.putIfAbsent(key, value)`

src/tools/debugger/session/Breakpoints.java
```diff
@@ -195,6 +203,33 @@ public Breakpoint getAsyncMessageRcvBreakpoint(final URI sourceUri,
...
+    }
+  }
+
+  public synchronized void updateReceiverBreakpoint(final BreakpointId bId, final Breakpoint bReceiver) {
+    if (receiverBreakpoints.keySet().contains(bId)) {
```
smarr: should that `if` better be an assert? I'd assume there is something broken if that condition does not hold, no?

src/tools/debugger/session/Breakpoints.java
ctrlpz: Right, I will fix it now

smarr:
I just pushed the change to Truffle and added the implementation using stepping.

src/tools/actors/ActorExecutionTrace.java
```diff
@@ -83,8 +79,8 @@ public static void recordMainActor(final Actor mainActor,
   /**
    * Assign breakpoint for the actor that is the receiver of the message.
    */
-  public static void assignBreakpoint(final Breakpoint breakpoint, final Actor actor,
-      final BreakpointDataTrace breakpointTrace, final boolean receiver) {
+  /*public static void assignBreakpoint(final Breakpoint breakpoint, final Actor actor,
```
ctrlpz: Probably it is best to delete these lines, this approach was based on REME-D implementation to save the breakpoints in the LocalManager (BreakpointActor in our case) we still have to figure out if we need this link between BreakpointActor and breakpoints.

src/tools/debugger/session/Breakpoints.java
```diff
@@ -195,6 +212,70 @@ public Breakpoint getAsyncMessageRcvBreakpoint(final URI sourceUri,
     return bp;
   }
 
+  public synchronized void addReceiverBreakpoint(final URI sourceUri, final int startLine,
+      final int startColumn, final int charLength) {
+    SectionBreakpoint bId = new SectionBreakpoint(sourceUri, startLine, startColumn, charLength);
```
ctrlpz: I think we can declare `BreakpointId bId = new SectionBreakpoint(sourceUri, startLine, startColumn, charLength);`
in order to use the abstract class

src/tools/actors/ActorExecutionTrace.java
smarr: done

src/tools/debugger/session/Breakpoints.java
smarr: Yeah, no, I prefer to have the concrete type here, especially since it goes into a map with breakpoints that is only supposed to contain receiver breakpoints. And, it makes the code using that map simpler.

PR #44, opened 2016-09-30
Remove duplicate code between web frontend and tests

smarr:
The goal is to finish the transition to TypeScript and remove code duplication where not necessary.

The changes should improve the overall code structure.
We are also going to move more code to a TypeScript style away from plain JavaScript.


tools/kompos/src/messages.ts
```diff
@@ -0,0 +1,207 @@
...
+  sourceSection: SourceSection;
+  methodName: string;
+}
+
+interface TopFrame {
+  arguments: string[];
+  slots:     IdMap<string>;
+}
+
+export type Message = SourceMessage | SuspendEventMessage | MessageHistoryMessage;
```
ctrlpz: What is the goal of it type of message?

tools/kompos/src/visualizations.ts
```diff
@@ -15,7 +15,7 @@ var width = 360,
 
 function hasSelfSends(actorId, messages) {
   for (var i in messages) {
-    if (messages[i].sender == "actorId") {
+    if (messages[i].sender === "actorId") {
       return true;
     }
   }
```
ctrlpz: What is this function for? Where it is used?

tools/kompos/src/messages.ts
smarr: Not sure what you are asking for. Do you want documentation? Things are going to change soon (including naming). So, I'd not yet invest the time on that.

tools/kompos/src/messages.ts
```diff
@@ -0,0 +1,207 @@
...
+  topFrame: TopFrame;
+  id: string;
+}
+
+export interface MessageHistoryMessage {
+  type: "messageHistory";
+  messageHistory: any; // TODO
+}
+
+export type BreakpointData = LineBreakpointData | SendBreakpointData | AsyncMethodRcvBreakpointData;
```
smarr: Stuff is going to change around soon again. So, I put them all in here for the moment.
There is also currently rather bad choice of terms with Message/Respond. That's very depending on what you consider sender/receiver. So, doesn't make sense universally. We'll need to find better terminology there, too.

tools/kompos/src/visualizations.ts
smarr: `visualizations.ts` is for graphical things, the actor history. While `view.ts` is for normal UI elements.

tools/kompos/src/messages.ts
ctrlpz: ok, but this code in message.ts file was first implemented in basic-protocol.ts, so the idea now is that it is not use only for testing, instead now Kompos uses those interfaces and the rest of your implementation and in basic-protocol.ts is only imported, right?

tools/kompos/src/messages.ts
ctrlpz: ok

tools/kompos/src/visualizations.ts
ctrlpz: ok thanks

tools/kompos/src/messages.ts
smarr: Yes, the idea is that there is only on set of type definitions for the tests and the web frontend.

PR #53, opened 2016-10-19
Intoduced Bugs to Savina

ad1622:
# dont accept/merge this pull request
- Savina.Philosophers: philosophers take a different fork than they put back (index off by one)
- Savina.PingPong: problem because the promise in the pong actor is resolved twice (2 pings one pong actor)
- Savina.Counting: counting is disturbed by another actor that increments when it is executed for the n/3 time => ~20-30% failure rate based on scheduling
- Savina.ThreadRing: missing decrement => endless loop
- Savina.Chameneos: missing messages lead to eternal waiting (actor causes color change but doesnt request another meeting)
- Savina.SleepingBarber: messed with message order in BarberActor enter, benchmark doesnt fail
- Savina.CigaretteSmokers: Smoker sends one too many startedsmoking messages to the arbitrator => benchmark doesnt fail but should be done faster as each smoke counts double.
- Savina.BankTransactions: all transactions are performed but the master is never notified about the transfered amounts => promise never resolved.


smarr:
This is something to remember for later experiments.
We should probably introduce that as an extra SavinaBuggy module or so.

PR #54, opened 2016-10-25
Update Actor Graph at Breakpoints

smarr:
- actor graph is now updated whenever a breakpoint is hit
- the data for the actor graph is sent in increments
- changed ObjectBuffer to allow iteration from a memorized point.
- added a clear method to ObjectBuffer

Reopend @ad1622's PR here to have CI and checks running, was before at https://github.com/MetaConc/SOMns/pull/1


smarr:
@ad1622 thanks!
@ctrlpz would be good if you have a look at the changes, to keep updated on what's going on.

PR #69, opened 2016-12-06
Explicit promise breakpoint

ctrlpz:
This pull request adds support for the following kinds of breakpoints:

- breakpoints set on `createPromisePair`, `whenResolved:`, `onError:`, `whenResolved:onError:`
- an asynchronous message breakpoint that halts after the message is executed. It uses `SteppingLocation.AFTER_STATEMENT` and the corresponding stepping strategy.

The main changes are:

- In `EventualMessage` class, there has been added two flags:
The flag `triggerMessageReceiverBreakpoint` indicates if an asynchronous message has a receiver breakpoint.
The flag `triggerPromiseResolverBreakpoint` indicates if a promise created implicitly has a resolver breakpoint.

- In `SPromise` class, there has been added four flags:
The flag `triggerPromiseResolutionBreakpoint` indicates  if a promise created implicitly or explicitly has a resolution breakpoint.
The flag `triggerResolutionBreakpointOnUnresolvedChainedPromise` indicates if an unresolved chained promise has a resolution breakpoint.
The flag `triggerExplicitPromiseResolverBreakpoint` indicates if a promise created explicitly has a resolver breakpoint.
The flag `explicitPromise` distinguishes implicit of explicit promises in the normal resolution of a promise (`ResolvePromiseNode`).

- Add a condition in the resolution of chained promises for the case when a promise resolution breakpoint is set on a promise with null resolver (`AbstractPromiseResolutionNode`).


src/som/interpreter/actors/SPromise.java
```diff
@@ -66,10 +66,19 @@ public static SPromise createPromise(final Actor owner) {
   // the owner of this promise, on which all call backs are scheduled
   protected final Actor owner;
 
-  protected SPromise(@NotNull final Actor owner) {
+  private boolean triggerResolutionBreakpointOnUnresolvedChainedPromise;
+  private boolean triggerPromiseResolverBreakpoint;
+  private boolean triggerPromiseResolutionBreakpoint;
+  private boolean explicitPromise;
```
smarr: These fields are not final? Do they need to change?

src/som/primitives/actors/PromisePrims.java
```diff
@@ -51,23 +60,46 @@ public boolean matches(final Object[] args, final ExpressionNode[] argNodes) {
              specializer = IsActorModule.class, noWrapper = true)
   public abstract static class CreatePromisePairPrim extends UnaryExpressionNode {
 
+    protected final AbstractBreakpointNode promiseResolverBreakpoint;
```
smarr: This can not work.
Truffle nodes need to be marked with the `@Child` annotation, like in the other places.
Also, the field cannot be marked as final, because the node needs to be able to change itself to a specialized version.

src/som/primitives/actors/PromisePrims.java
```diff
@@ -51,23 +60,46 @@ public boolean matches(final Object[] args, final ExpressionNode[] argNodes) {
...
     }
 
-    public CreatePromisePairPrim(final boolean eagerWrapper, final SourceSection source) { super(eagerWrapper, source); }
+    public CreatePromisePairPrim(final boolean eagerWrapper, final SourceSection source) {
+      super(eagerWrapper, source);
+
+      if (VmSettings.TRUFFLE_DEBUGGER_ENABLED) {
+        Breakpoints breakpointCatalog = VM.getWebDebugger().getBreakpoints();
+        FullSourceCoordinate sourceCoord = SourceCoordinate.create(source);
+        this.promiseResolverBreakpoint   = insert(BreakpointNodeGen.create(breakpointCatalog.getPromiseResolverBreakpoint(sourceCoord)));
```
smarr: since we have multiple places for this stuff, I wonder whether it wouldn't be better to have a helper method for the breakpoint node creation. A factory method, that internally checks the `VmSettings. TRUFFLE_DEBUGGER_ENABLED` flag.

src/som/primitives/actors/PromisePrims.java
```diff
@@ -184,7 +216,7 @@ protected static final SPromise whenResolvedOrError(final SPromise rcvr,
       assert error.getMethod().getNumberOfArguments() == 2;
 
       Actor current = EventualMessage.getActorCurrentMessageIsExecutionOn();
-      SPromise  promise  = SPromise.createPromise(current);
+      SPromise  promise  = SPromise.createPromise(current, false, false, true);
```
smarr: all the places where you pass predefined sets of `false` flags, are likely place where one might want to be able to set a breakpoint.
So, here, I think, we are at the place where the promise that is returned by a `whenResolved:`  is created.
So, actually, we probably would like to have promise breakpoints on `whenResolved:`, too.

src/tools/actors/Tags.java
```diff
@@ -7,4 +7,8 @@
 
   }
 
+  public final class CreatePromisePairTag extends Tags{
```
smarr: `Tags{` --> `Tags {` (whitespace)

tools/kompos/src/debugger.ts
```diff
@@ -2,10 +2,14 @@ import {IdMap, Source, SourceCoordinate, SourceMessage, TaggedSourceCoordinate,
   getSectionId} from './messages';
 import {Breakpoint} from './breakpoints';
 
-export function isRelevant(sc: TaggedSourceCoordinate) {
+export function isRelevantSendTag(sc: TaggedSourceCoordinate) {
```
smarr: this is not proper english. Now the name needs to be started with `has` instead of `is`.

tools/kompos/tests/basic-protocol.ts
```diff
@@ -451,6 +451,78 @@ describe('Basic Protocol', function() {
      }));
   });
 
+   describe('setting a source section promise resolution breakpoint on unresolved explicit promise', () => {
```
smarr: Indentation for the whole code snippet added here looks wrong.

src/som/primitives/actors/PromisePrims.java
ctrlpz: Why in the abstract nodes that are in the SendNode class there is not @Child annotation and in that case are final?

ctrlpz: ok, I will fix it.

src/som/primitives/actors/PromisePrims.java
ctrlpz: the idea here will be to propagate the flags that has `rcvr` promise?

smarr: maybe that, but more importantly, we want to be able to have a popup there, and set them explicitly.

src/som/compiler/Parser.java
```diff
@@ -1126,8 +1132,16 @@ protected ExpressionNode unaryMessage(final ExpressionNode receiver,
       final boolean eventualSend, final SourceSection sendOperator) throws ParseError {
     SourceCoordinate coord = getCoordinate();
     SSymbol selector = unarySelector();
-    return createMessageSend(selector, new ExpressionNode[] {receiver},
-        eventualSend, getSource(coord), sendOperator);
+
+    if (selector.getString().equals("createPromisePair")) {
```
smarr: this is now obsolete. See https://github.com/smarr/SOMns/pull/88. The already merged pull requested does that for you automatically and for all other primitives.

src/tools/debugger/session/Breakpoints.java
```diff
@@ -74,6 +74,7 @@ public synchronized void addOrUpdate(final LineBreakpoint bId) {
       WebDebugger.log("LineBreakpoint: " + bId);
       bp = Breakpoint.newBuilder(bId.getURI()).
           lineIs(bId.getLine()).
+          steppingLocation(SteppingLocation.BEFORE_STATEMENT).
```
smarr: Not necessary with my latest truffle changes. (same below, I believe)

src/tools/debugger/session/Breakpoints.java
```diff
@@ -82,11 +83,16 @@ public synchronized void addOrUpdate(final LineBreakpoint bId) {
...
 
-  public synchronized void addOrUpdate(final AsyncMessageReceiverBreakpoint bId) {
-    Breakpoint bp = saveTruffleBasedBreakpoints(bId, RootTag.class);
+  public synchronized void addOrUpdate(final AsyncMessageBeforeExecutionBreakpoint bId) {
+    Breakpoint bp = saveTruffleBasedBreakpoints(bId, RootTag.class, SteppingLocation.BEFORE_STATEMENT);
+    bp.setCondition(BreakWhenActivatedByAsyncMessage.INSTANCE);
+  }
+
+  public synchronized void addOrUpdate(final AsyncMessageAfterExecutionBreakpoint bId) {
+    Breakpoint bp = saveTruffleBasedBreakpoints(bId, RootTag.class, SteppingLocation.AFTER_STATEMENT);
```
smarr: This might be the only one that's strictly necessary.

tools/kompos/index.html
```diff
@@ -8,7 +8,7 @@
   <script src="/node_modules/jquery/dist/jquery.js"></script>
   <script src="/out/src/libs/jquery-resizable.js"></script>
   <script src="/node_modules/tether/dist/js/tether.js"></script>
-  <script src="/node_modules/bootstrap/dist/js/bootstrap.js"></script>
+  <script src="/node_modules/bootstrap/dist/js/bootstrap.js"></script>
```
smarr: whitespace change, why?

tools/kompos/index.html
```diff
@@ -155,20 +167,18 @@
     font: 10px sans-serif;
   }
 
-  #header  {padding-left: 3em; padding-top: 1ex}
+ #header  {padding-left: 3em; padding-top: 1ex}
```
smarr: whitespace change, why?

tools/kompos/index.html
```diff
@@ -211,14 +221,13 @@
...
   }
-
   .activity-scopes {
     width: 100%;
     margin-top: 2ex;
   }
   .activity-scopes th, .activity-scopes td {
     padding: 0.1rem .2rem;
-    font-size: 9px
+    font-size: 12px
```
smarr: please don't change these visual things as part of this pull request.
They have nothing to do with breakpoints.

tools/kompos/src/debugger.ts
```diff
@@ -8,6 +8,14 @@ export function isRelevant(sc: TaggedSourceCoordinate) {
   return -1 !== sc.tags.indexOf("ExpressionBreakpoint");
 }
 
+export function isRelevantPair(sc: TaggedSourceCoordinate) { 
+  return -1 !== sc.tags.indexOf('CreatePromisePair');
```
smarr: Perhaps you could just include the `ExpressionBreakpoint` tag on those nodes as well. It seems strange to have extra tests here, while the other breakpoints unified this nicely.

tools/kompos/src/debugger.ts
```diff
@@ -69,9 +77,9 @@ export class Debugger {
     let sId = this.getSourceId(s.uri);
     for (let sc of s.sections) {
       // Filter out all non-relevant source sections
-      if (isRelevant(sc)) {
+      if (isRelevant(sc) || isRelevantPair(sc) || isRelevantWhenResolved(sc)) {      
         let id = getSectionId(sId, sc);
-        this.sections[id] = sc;
+        this.sections[id] = sc;
```
smarr: Why the whitespace change?

tools/kompos/src/view.ts
```diff
@@ -388,13 +388,72 @@ function constructChannelBpMenu(fileNode, tag: string, tpl: string) {
   });
 }
 
+function enableCreatePromisePairClicks(fileNode) {
```
smarr: a lot of code duplication introduce here, could that please be avoided, also for existing code?

tools/kompos/tests/basic-protocol.ts
```diff
@@ -190,7 +239,7 @@ describe("Basic Protocol", function() {
     });
   });
 
-  describe("stepping", () => {
+  describe('stepping', () => {
```
smarr: why this change?

src/som/interpreter/actors/AbstractPromiseResolutionNode.java
```diff
@@ -5,18 +5,29 @@
...
 
   @Child protected WrapReferenceNode wrapper = WrapReferenceNodeGen.create();
+  @Child protected UnaryExpressionNode haltNode;
 
-  protected AbstractPromiseResolutionNode(final boolean eagWrap, final SourceSection source) { super(eagWrap, source); }
-  protected AbstractPromiseResolutionNode(final AbstractPromiseResolutionNode node) { super(node); }
+  protected AbstractPromiseResolutionNode(final boolean eagWrap, final SourceSection source) {
+    super(eagWrap, source);
+    haltNode = insert(SuspendExecutionNodeGen.create(false, source, 2, null));
+    VM.insertInstrumentationWrapper(haltNode);
```
ctrlpz: the initialization of the `haltNode` is required, it is used in `ResolvePromiseNode.normalResolution` method. If I remove this initialization I would have a `NullPointerException` when trying to set a resolver breakpoint in an explicit promise.

smarr: Ehm, I am not questioning the initialization of the `haltNode`.
I am just wondering whether `haltNode = SuspendExecutionNodeGen.create(false, source, 2, null);` would be sufficient.

src/som/primitives/actors/PromisePrims.java
```diff
@@ -148,11 +224,11 @@ protected static final SPromise registerOnError(final SPromise rcvr,
       assert block.getMethod().getNumberOfArguments() == 2;
 
       Actor current = EventualMessage.getActorCurrentMessageIsExecutionOn();
-      SPromise  promise  = SPromise.createPromise(current);
+      SPromise  promise  = SPromise.createPromise(current, false, false, false);
```
ctrlpz: Right now there is no implementation for putting breakpoints on `onError` primitive and `actorsWhen:resolved:onError`, that's why I set leave them with false value, I will include them

src/tools/debugger/session/Breakpoints.java
ctrlpz: which changes are you refering to? I only see this one as last commit https://github.com/smarr/truffle/commit/202248be40c6166a49ea664b8b2ef3e4e8980f96

smarr: This one: https://github.com/smarr/truffle/commit/a4b4d2b942578d9b8b2bd9b19fb9e5836b77703f#diff-6be44cfcbe543486a933eb106a501f8aR775

smarr:
@ctrlpz I polished the PR for merging. Originally, the history was super confusing with all the merge/rebase code reverts and restores. I squashes everything into one commit and then cleaned up code duplication, also improving the general code elsewhere a little.

Before it can be merged, please check the following things:
- [ ] there is a TODO comment, which still seems like it needs to be processes
- [ ] the Kompos interface needs to be tested manually, would be great if you could have a look
- [ ] the flag that's duplicated in the promise and the message, would be good to see whether we could only have one
- [ ] please clarify the PR description to be understandable without a lot of background

Thanks!

src/som/interpreter/actors/AbstractPromiseResolutionNode.java
```diff
@@ -51,15 +63,19 @@ protected void chainPromise(final SResolver resolver,
...
     synchronized (promiseValue) {
       Resolution state = promiseValue.getResolutionStateUnsync();
       if (SPromise.isCompleted(state)) {
         resolvePromise(state, resolver, promiseValue.getValueUnsync(),
-           isBreakpointOnPromiseResolution);
+            breakpointOnResolution);
       } else {
         synchronized (promiseToBeResolved) { // TODO: is this really deadlock free?
-          promiseToBeResolved.setTriggerResolutionBreakpointOnUnresolvedChainedPromise(isBreakpointOnPromiseResolution);
+          // TODO: check this condition.
```
smarr: @ctrlpz you are saying I should check this condition?
I don't know what you want me to do here. The condition looks ok, but I have no idea what you were thinking when you added the todo. What circumstances are you not sure about? What situation might arise where this is problematic?

This TODO note doesn't tell me how I can help you here.

ctrlpz: if I don't add that condition there is a case where the flag isBreakpointOnResolution is not propagated, look at this conversation for more detail https://metaconc.slack.com/archives/C1ZK5N4FQ/p1489677673962877

smarr: ok, but what do you want me to check here? Is there anything else that needs to be considered? Why is there a TODO here?

ctrlpz: no, apart from the situation I described in the link I don't think there is another case where the flag is false. I just had the doubt why in the case that the resolver is null, the promiseValue is the one that has the flag true? If you consider that the condition is correct then please remove the TODO.

Issue #48, opened 2016-10-10
Add a timeout primitive for executing blocks after a set time interval


smarr:
Should probably look something like this in the actor module:

``` Smalltalk
(* Executes aBlock after waiting for `milliseconds`.
   Returns a promise, which is going to be resolved to the
   return value of the block. *)
public after: milliseconds do: aBlock = (
  ^ vmMirror actorsDo: aBlock after: milliseconds
)
```

smarr:
To implement this, the simplest strategy is probably to create a promise pair in the SOM side, and pass the timeout and the resolve to the vmMirror.
The promise should get the block as a `whenResolved:`.

That way, we ensure that the block is execute on the right actor without extra complications and without having to worry that the timer is racing with the event loop.

smarr:
Solved with #70.

PR #70, opened 2016-12-07
Implemented timeout-primitive

ad1622:
Implemented a timeout-primitive that allows to execute a block after a specified number of ms passed. #48

### DIFF
--- a/src/som/primitives/TimerPrim.java
+++ b/src/som/primitives/TimerPrim.java
@@ -1,0 +1,47 @@
+package som.primitives;
+
+import java.util.Timer;
+import java.util.TimerTask;
+
+import com.oracle.truffle.api.dsl.GenerateNodeFactory;
+import com.oracle.truffle.api.dsl.Specialization;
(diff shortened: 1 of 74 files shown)


src/som/primitives/TimerPrim.java
```diff
@@ -0,0 +1,47 @@
...
+@Primitive(primitive = "actorResolveProm:after:")
+public abstract class TimerPrim extends BinaryComplexOperation{
+  protected TimerPrim(final BinaryComplexOperation node) { super(node); }
+  protected TimerPrim(final boolean eagWrap, final SourceSection source) { super(eagWrap, source); }
+
+  @Child protected WrapReferenceNode wrapper = WrapReferenceNodeGen.create();
+
+  @Specialization
+  public final Object doResolveAfter(final VirtualFrame frame, final SResolver resolver, final long timeout) {
+    Timer t = new Timer();
```
smarr: just seeing this one. I think, I am going to change that before merging to be a field of the node. Or perhaps even a static field. We don't want a separate timer thread for each time out :)

PR #71, opened 2016-12-07
Added Flag to disable writing trace

ad1622:


som
```diff
@@ -73,6 +73,8 @@ tools.add_argument('-mt', '--memory-tracing', help='enable tracing of memory usa
                     dest='memory_tracing', action='store_true', default=False)
 parser.add_argument('-tf', '--trace-file', help='Trace file destination, default: traces/trace',
                     dest='trace_file', default=BASE_DIR + '/traces/tracea')
+tools.add_argument('-dtf', '--disable-trace-file', help='trace file wont be written to disk',
```
smarr: minor change request, flags that disable stuff have been capitalized like `-G`. to stay consistent, I would propose to use `-TF`

src/tools/actors/ActorExecutionTrace.java
```diff
@@ -370,6 +370,7 @@ public void run() {
       File sf = new File(VmSettings.TRACE_FILE + ".sym");
       f.getParentFile().mkdirs();
 
+
```
smarr: superfluous newline?

PR #74, opened 2016-12-14
Catch exceptions for implicit promises

ctrlpz:
- change resolution of the promise to ERRORNOUS when occurs an exception
- set the value of the SomException as the returned value of the promise message.
- TODO: implement resolver error for explicit promises

In the program TestExceptionPromises of small-experiments it is captured an exception thrown by an implicit promise.
In this branch is needed still to consider the cases when an exception is thrown for promises that are declared explicitly. 

There are two more TODO but related to the branch explicit-promise-breakpoints (check pull request https://github.com/smarr/SOMns/pull/69)
1.  Solve issue of double wrapping of the primitive whenResolved, it causes assertion error in method isTaggedWith of EagerBinaryPrimitiveNode class.
2. There is an issue related to the the tests of breakpoints. This happen after the  changes of the promise breakpoint flags. All tests in basic-protocol.ts passed except for SOMns stack trace test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ newspeak
 tools/kompos/node_modules
 tools/kompos/out
 traces
+/npm-debug.log
--- a/src/som/interpreter/actors/ReceivedMessage.java
+++ b/src/som/interpreter/actors/ReceivedMessage.java
@@ -9,6 +9,7 @@ import com.oracle.truffle.api.nodes.DirectCallNode;
 
 import som.VmSettings;
 import som.interpreter.SArguments;
+import som.interpreter.SomException;
 import som.interpreter.SomLanguage;
 import som.interpreter.nodes.MessageSendNode.AbstractMessageSendNode;
 import som.vmobjects.SSymbol;
@@ -35,9 +36,12 @@ public class ReceivedMessage extends ReceivedRootNode {
       dbg.prepareSteppingAfterNextRootNode();
     }
 
-    Object result = onReceive.doPreEvaluated(frame, msg.args);
-
-    resolvePromise(frame, msg.resolver, result, msg.triggerPromiseResolutionBreakpoint);
+    try {
+      Object result = onReceive.doPreEvaluated(frame, msg.args);
+      resolvePromise(frame, msg.resolver, result, msg.triggerPromiseResolutionBreakpoint);
+    } catch (SomException exception) {
+      msg.getResolver().onError(exception);
+    }
     return null;
   }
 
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -14,3 +14,4 @@ newspeak
 tools/kompos/node_modules
 tools/kompos/out
 traces
+/npm-debug.log
```
smarr: Please remove the leading `/`, then it matches `npm-debug.log` files everywhere I think, which would be a bit more useful.

src/som/interpreter/actors/ReceivedMessage.java
```diff
@@ -35,9 +36,12 @@ public Object execute(final VirtualFrame frame) {
...
     }
 
-    Object result = onReceive.doPreEvaluated(frame, msg.args);
-
-    resolvePromise(frame, msg.resolver, result, msg.triggerPromiseResolutionBreakpoint);
+    try {
+      Object result = onReceive.doPreEvaluated(frame, msg.args);
+      resolvePromise(frame, msg.resolver, result, msg.triggerPromiseResolutionBreakpoint);
+    } catch (SomException exception) {
+      msg.getResolver().onError(exception);
```
smarr: I think I would prefer that you do `exception.getSomObject()` here. So, that `onError` does not need to deal with this special case.

smarr: Also, using `onError` directly here seems incorrect. Looking at the implementation of `onError()`, it is definitely incorrect.

The error case needs to be resolved very similarly to the normal case.
So, I'd expect the same code structure, using something like `resolvePromise(...)` from the normal case, too.
`resolvePromise()` uses the `ResolvePromiseNode` which does synchronization and handles when promises are already resolved, etc. All this, needs also to be done for the error case.

src/som/interpreter/actors/ReceivedMessage.java
```diff
@@ -85,9 +89,12 @@ public Object execute(final VirtualFrame frame) {
...
 
-      Object result = onReceive.call(frame, msg.args);
-
-      resolvePromise(frame, msg.resolver, result, msg.triggerPromiseResolutionBreakpoint);
+      try {
+        Object result = onReceive.call(frame, msg.args);
+        resolvePromise(frame, msg.resolver, result, msg.triggerPromiseResolutionBreakpoint);
+      } catch (SomException exception) {
+        msg.getResolver().onError(exception);
+      }
```
smarr: Same comments as above.
Might be useful to thing about how to remove that code duplication, too.

src/som/interpreter/actors/SPromise.java
```diff
@@ -141,9 +141,6 @@ private void registerMoreWhenResolved(final PromiseMessage msg) {
 
   public synchronized void registerOnError(final PromiseMessage msg,
       final Actor current) {
-    if (resolutionState == Resolution.ERRORNOUS) {
-      scheduleCallbacksOnResolution(value, msg, current, false);
-    } else {
```
smarr: why is this code removed? I don't understand?

smarr:
Should be superseded by #118.

PR #77, opened 2016-12-19
Added  actor tracing configuration flag

ad1622:
use -atcfg=<config> to turn off tracing for some things, for example -atcfg=mt:mp:pc to disable message timestamps, message parameters and promises.

### DIFF
--- a/src/som/VmSettings.java
+++ b/src/som/VmSettings.java
@@ -1,5 +1,7 @@
 package som;
 
+import java.util.Arrays;
+import java.util.List;
 
 public class VmSettings {
   public static final int NUM_THREADS;
@@ -13,6 +15,11 @@ public class VmSettings {
   public static final boolean INSTRUMENTATION;
   public static final boolean DYNAMIC_METRICS;
   public static final boolean DNU_PRINT_STACK_TRACE;
+  public static final boolean MESSAGE_TIMESTAMPS;
+  public static final boolean MESSAGE_PARAMETERS;
+  public static final boolean PROMISE_CREATION;
+  public static final boolean PROMISE_RESOLUTION;
+  public static final boolean PROMISE_RESOLVED_WITH;
 
   public static final boolean TRUFFLE_DEBUGGER_ENABLED;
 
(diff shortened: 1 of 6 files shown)


src/tools/actors/ActorExecutionTrace.java
```diff
@@ -275,7 +279,17 @@ public static void mailboxExecuted(final Mailbox m, final Actor actor) {
       b.put(Events.Mailbox.id);
       b.putLong(m.getBasemessageId()); //base id for messages
       b.putLong(actor.getActorId()); //receiver of the messages
+
       int idx = 0;
+      byte eventid = MESSAGE_BASE;
+
+      if (VmSettings.MESSAGE_TIMESTAMPS) {
```
smarr: ok, you have to explain to me why this bit is set depending on the static configuration of the interpreter. This is not depending on a dynamic value. So, why is this info relevant in each event id?

(same comment below)

src/tools/actors/ActorExecutionTrace.java
```diff
@@ -287,32 +301,36 @@ public static void mailboxExecuted(final Mailbox m, final Actor actor) {
           b.putShort((short) idx);
         }
 
-        if (em instanceof PromiseSendMessage) {
-          b.put(Events.PromiseMessage.id);
+        if (em instanceof PromiseSendMessage && VmSettings.PROMISE_CREATION) {
+          b.put((byte) (eventid | PROMISE_BIT));
```
smarr: here it depends on run time info, but, it seems this bit wasn't there before. So, why is it needed now?

src/som/VmSettings.java
```diff
@@ -33,6 +40,16 @@
     MEMORY_TRACING = getBool("som.memoryTracing",   false);
     DISABLE_TRACE_FILE = getBool("som.disableTraceFile", false);
 
+    //disabling of message or actor tracing in general is probably not a good idea and would require to turn off other tracing as well;
```
ad1622: The comment was intended to say, that turning off actor/message tracing would be as useful as no tracing at all.

src/tools/actors/ActorExecutionTrace.java
ad1622: when promise tracing is turned off, the promise message is treated like a normal message

PR #90, opened 2017-01-16
Optimize Single Message Processing in Mailbox

daumayr:
Replaces the current Mailbox implementation with a approach similar to how SPromise handles chained Promises. This brings a performance improvement due to currently low mailbox utilization (many contain only one message).


src/som/interpreter/actors/Actor.java
```diff
@@ -195,44 +220,74 @@ public void run() {
...
+
+        current.execute();
         if (VmSettings.ACTOR_TRACING) {
           currentThread.currentMessageId += 1;
         }
+
+        int i = 0;
+        if (size > 1) {
+          for (EventualMessage msg : moreCurrent) {
+            actor.logMessageBeingExecuted(msg);
```
smarr: this block looks identical to the handling of the first message. could this be a helper method?

PR #97, opened 2017-01-24
Added Peak Memory Report

daumayr:
running som with "-mt" flag reports peak memory usage for heap and non heap memory, additionally  the time spent on GC is provided.
Data is collected from MXBeans just before the VM shuts down.

example:
[Memstat] Heap: 209142184B	NonHeap: 44378016B	Collected: 518722336B	GC-Time: 67ms

### DIFF
--- a/src/som/VM.java
+++ b/src/som/VM.java
@@ -180,6 +180,9 @@ public final class VM {
     }
     Actor.shutDownActorPool();
     engine.dispose();
+    if (VmSettings.MEMORY_TRACING) {
+      ActorExecutionTrace.reportPeakMemoryUsage();
+    }
     System.exit(errorCode);
   }
 
--- a/src/tools/concurrency/ActorExecutionTrace.java
+++ b/src/tools/concurrency/ActorExecutionTrace.java
@@ -8,13 +8,14 @@ import java.io.IOException;
 import java.io.OutputStreamWriter;
 import java.lang.management.GarbageCollectorMXBean;
 import java.lang.management.ManagementFactory;
-import java.lang.management.MemoryMXBean;
 import java.lang.management.MemoryPoolMXBean;
 import java.lang.management.MemoryType;
+import java.lang.management.MemoryUsage;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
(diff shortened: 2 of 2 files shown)


src/tools/actors/ActorExecutionTrace.java
```diff
@@ -46,6 +43,7 @@
   private static final byte PROMISE_BIT = 0x40;
   private static final byte TIMESTAMP_BIT = 0x20;
   private static final byte PARAMETER_BIT = 0x10;
+  private static final int MB_FACTOR = 1000000;
```
smarr: I guess that's MB with MiB? Somewhat unexpected.

src/tools/actors/ActorExecutionTrace.java
```diff
@@ -93,32 +87,23 @@ public static void recordMainActor(final Actor mainActor,
...
-          if (notification.getType().equals(GarbageCollectionNotificationInfo.GARBAGE_COLLECTION_NOTIFICATION)) {
-            // get the information associated with this notification
-            GarbageCollectionNotificationInfo info = GarbageCollectionNotificationInfo.from((CompositeData) notification.getUserData());
-            VM.println(Thread.currentThread().toString());
-            VM.println(info.getGcAction() + ": - " + info.getGcInfo().getId() + " " + info.getGcName() + " (from " + info.getGcCause() + ") " + info.getGcInfo().getDuration() + " ms;");
-            VM.println("GcInfo MemoryUsageBeforeGc: " + info.getGcInfo().getMemoryUsageBeforeGc().entrySet().stream().filter(ent -> !ent.getKey().equals("Compressed Class Space") && !ent.getKey().equals("Code Cache")).mapToLong(usage -> usage.getValue().getUsed()).sum() / 1024 + " kB");
-            VM.println("GcInfo MemoryUsageAfterGc: " + info.getGcInfo().getMemoryUsageAfterGc().entrySet().stream().filter(ent -> !ent.getKey().equals("Compressed Class Space") && !ent.getKey().equals("Code Cache")).mapToLong(usage -> usage.getValue().getUsed()).sum() / 1024 + " kB");
-          }
-        }
-      };
```
smarr: What's the code delated here?
You were tracking memory usage before and after GC?
If it is easy and low overhead, I think it would be good to also have a number about how much memory was collected over time.
That would mean, we could have a number of how much memory was allocated in total, right?

Issue #83, opened 2017-01-05
Add support for literal arrays


smarr:
I really miss the ability to simply note down a literal array.

As far as I can see, the [spec](http://bracha.org/newspeak-spec.pdf) only talks about arrays in very few places, and uses the `{2. 4. 5}` notation. That is a list of expressions, of which their return value defines the value of the corresponding array index.
So, `{self foo. 4 + 3. a <-: hello: 'ww'}` is perfectly legal, and might evaluate to `{self. 7. a Promise}`.

What's unclear is how we can have value arrays supported as well.
Perhaps one option would be to use the `#` sign as an indicator, and make `#{2. 4. 5}` a value array. That's somewhat in Smalltalk tradition of `#(foo bar 42)` evaluating to `{#foo. #bar. 42}`, except that Smalltalk doesn't have `ValueArray`s, and this seems merely a convenience to note down a list of symbols without too much syntactical noise. Not sure I really care for such symbol arrays.

fniephaus:
How do you currently define a value array? Can you do something like `{2. 4. 5} asValueArray`?

smarr:
There is no such helper in SOMns.
There is only this: https://github.com/smarr/SOMns/blob/master/core-lib/Kernel.som#L572

So, it would be `i := 0. ValueArray new: arr size withAll: [ i := i + 1. arr at: i ]`.

And a brief grep reveals that I don't even use ValueArrays currently. I added them for one of the Savina benchmarks, but, seems I never completed the implementation of that benchmark.

smarr:
Status update: this task is worked on in #100

smarr:
With #173 merged, we got array literals.
However, I'll leave the issue here open, since I'd still love to have literal value arrays in the form of `#{#44. 55. MyValue new: 55}`, which guarantee to be deeply immutable.

PR #100, opened 2017-01-27
Implemented array literals

richard-roberts:
Add support for literal arrays, based on issue #83

In the Parser, I've added a new primary expression. When the literal array opening symbol `{` is found, a new method `literalArray(builder) ` is called. This method parses each expression uses a while loop, much like is done in `blockBody(builder)`. The method also eats the braces and the periods between expressions. Once collected, the list of expressions is sent to a new method in the node factory, which marks each expression as a statement and then returns an instance of the new Array Literal node. When executed, the Array Literal node instantiates an SMutableArray by evaluating each expressions.

Some things I'd like to do next are:

- [ ] optimization the Array Literal node using Truffle specializations
- [ ] write appropriate tests to validate that the implementation works as expect
- [ ] implement support for value arrays (as noted in the issue linked above)

### DIFF
--- a/src/som/compiler/Parser.java
+++ b/src/som/compiler/Parser.java
@@ -35,6 +35,7 @@ import static som.compiler.Symbol.Div;
 import static som.compiler.Symbol.Double;
 import static som.compiler.Symbol.EndBlock;
 import static som.compiler.Symbol.EndComment;
+import static som.compiler.Symbol.EndLiteralArray;
 import static som.compiler.Symbol.EndTerm;
 import static som.compiler.Symbol.Equal;
 import static som.compiler.Symbol.EventualSend;
@@ -50,6 +51,7 @@ import static som.compiler.Symbol.Mod;
 import static som.compiler.Symbol.More;
 import static som.compiler.Symbol.NONE;
 import static som.compiler.Symbol.NewBlock;
+import static som.compiler.Symbol.NewLiteralArray;
 import static som.compiler.Symbol.NewTerm;
 import static som.compiler.Symbol.Not;
 import static som.compiler.Symbol.OperatorSequence;
@@ -62,6 +64,7 @@ import static som.compiler.Symbol.STString;
 import static som.compiler.Symbol.SlotMutableAssign;
 import static som.compiler.Symbol.Star;
 import static som.interpreter.SNodeFactory.createImplicitReceiverSend;
+import static som.interpreter.SNodeFactory.createLiteralArray;
 import static som.interpreter.SNodeFactory.createMessageSend;
 import static som.interpreter.SNodeFactory.createSequence;
 import static som.vm.Symbols.symbolFor;
@@ -1040,6 +1043,9 @@ public class Parser {
           return new BlockNode(blockMethod, lastMethodsSourceSection);
         }
       }
+      case NewLiteralArray: {
+        return literalArray(builder);
+      }
       default: {
         if (symIn(literalSyms)) {
           return literal();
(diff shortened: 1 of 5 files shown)


src/som/compiler/Parser.java
```diff
@@ -1444,6 +1450,31 @@ private LiteralNode literalString() throws ParseError {
...
+        return createLiteralArray(expressions, getSource(coord));
+      }
+
+      if (!sawPeriod) {
+        expect(Period, null);
+      }
+
+      expressions.add(expression(builder));
+      sawPeriod = accept(Period, StatementSeparatorTag.class);
+    }
```
smarr: Hm, I wonder, could that not be just:

```Java
do {
  comments();
  expressions.add(expression(builder));
} while (accept(Period...));

comments();
expect(EndLiteralArray, DelimiterClosingTag.class);
return createLiteralArray(expressions, getSource(coord));
```

richard-roberts: Hmm good idea, it would be nice to make the method more concise. However, unless I've misunderstood, I believe that this suggestion would throw an error on an empty literal array: `{}`. One option would be to check for `EndLiteralArray` before the loop, which should work.

Is there anything that can go into the literal array that won't parse as an expression? If not, then I will go ahead and make the change!

smarr: You're right, I missed that case.

Comments and expressions are the only thing that's supported by the grammar.

richard-roberts: I restructured the while loop, which is now more concise using a switch statement. The current implementation expects the user to properly provide expressions, with periods delimiting them (with the last being optional). 

Currently there is nothing done to communicate why the program's definition is invalid. Should I add something to accompany the `Program Definition Error`in cases where a parsable expression cannot be found (or will this be handled sufficiently in parse expression)?

src/som/compiler/Parser.java
```diff
@@ -1444,6 +1450,29 @@ private LiteralNode literalString() throws ParseError {
...
+    expect(NewLiteralArray, DelimiterOpeningTag.class);
+
+    while (true) {
+      comments();
+
+      switch (sym) {
+        case EndLiteralArray:
+          expect(EndLiteralArray, DelimiterClosingTag.class);
+          return createLiteralArray(expressions, getSource(coord));
+        case Period:
```
smarr: hmm, I think this is not entirely correct.

This is more like: `{ DOT* expr*  }`.

This means it parses `{....3....}` and it is `{3}`, which is surprising, I think.

And, it also parses `{3 3 3 3}`, which is `{3. 3. 3. 3}`. This should also not be supported.

richard-roberts: Yeah, that's true - I got carried away trying to make the loop more concise. So perhaps stay with the original version in that case?

smarr: Yeah, I couldn't come up with a better solution either.

But, I'd prefer to change the naming of the bool, and flip its value, roughly like this:

```Java
boolean needsSeparator = false;
while (true) {
  comments();
  if (sym == EndLiteralArray) {
    expect(EndLiteralArray, DelimiterClosingTag.class);
    return createLiteralArray(expressions, getSource(coord));
  }
  if (needsSeparator) {
    expect(Period, null);
  }
  expressions.add(expression(builder));
  needsSeparator = !accept(Period, StatementSeparatorTag.class);
}
```

When it comes to nicer error messages, you can always use the `expect(.)`/`accept(.)` methods that take a custom message.

richard-roberts: Agreed, that suggestion is much more clear. I've made the change, although when you say _roughly_ perhaps you meant _exactly_.

smarr:
Hi Richard, does this mean you're not going to complete this? Just wondering.

richard-roberts:
Hi Stefan, sorry for the confusion - I do intend to complete this! I inadvertently merged (and deleted) the branch into a new _Grace_ development branch in my fork, which contains both the implementation of array and object literals. Is this okay? I'm happy to recreate the branch if you would prefer them as seperate entities.

smarr:
Don't worry at it for the moment. Yes, it would be great to keep array and object literals separate, just to make it easier to track where the code comes from, and what considerations went into it.
So, having separate pull requests helps.
But, we can deal with that once you got things in a state that you like.

(another thing is that reviewing a huge PR is hard, and important things might be missed, not just code issues, but also performance related things)

richard-roberts:
Okay, I understand ~ I will fix this up once I've stabilized the implementation. Thanks!

smarr:
@richard-roberts can I adopt this patch? Would like to get array literals for Newspeak compatibility.

PR #109, opened 2017-02-28
Implemented trace replay

daumayr:
Replay feature enforces the message order recorded in the trace file. Messages that don't fit the next expected message of an actor are put into a leftover queue. After a fitting message was received, leftovers and remaining mailbox are checked for other messages that can be executed. The leftovers have a higher priority. Extra messages that are not present in the trace will not be processed (e.g. weren't processed due to shutdown in original run).

Actors that didn't exist in the trace cause an assertion exception.

Replay requires the program parameters to be equal to those of the execution recorded in the trace.
The easiest way to replay, is to use the same command and adding the replay flag `-r`.

- fixed tracefile bugs
- added a mailbox number to the mailbox-events to allow ordering.
- added `-vmd` flag to enable debug output.
- added `-r` flag to enable replay.
- added `-J` flag for JVM flags, e.g. `-JXmx2g`
- removed `DebugActor`, tracing provides the same and more functionality

### DIFF
--- a/src/som/interpreter/actors/Actor.java
+++ b/src/som/interpreter/actors/Actor.java
@@ -54,9 +54,7 @@ import tools.debugger.WebDebugger;
 public class Actor implements Activity {
 
   public static Actor createActor() {
-    if (VmSettings.DEBUG_MODE) {
-      return new DebugActor();
-    } else if (VmSettings.REPLAY) {
+    if (VmSettings.REPLAY) {
       return new ReplayActor();
     } else if (VmSettings.ACTOR_TRACING) {
       return new TracingActor();
@@ -116,13 +114,19 @@ public class Actor implements Activity {
     executor = new ExecAllMessages(this);
   }
 
-  public static void printMissingMessages() {
+  /**
+   * Prints a list of expected Messages and remaining mailbox content.
+   * @return true if there are actors expecting messages, false otherwise.
+   */
+  public static boolean printMissingMessages() {
     if (!(VmSettings.REPLAY && VmSettings.DEBUG_MODE)) {
-      return;
+      return false;
     }
 
+    boolean result = false;
     for (Actor a : actorList) {
       if (a.getReplayExpectedMessages() != null && a.getReplayExpectedMessages().peek() != null) {
+        result = true; // program did not execute all messages
         if (a.getReplayExpectedMessages().peek() instanceof TraceParser.PromiseMessage) {
           VM.println(a.getName() + " [" + a.getReplayActorId() + "] expecting PromiseMessage " + a.getReplayExpectedMessages().peek().symbol + " from " + a.getReplayExpectedMessages().peek().sender + " PID " + ((TraceParser.PromiseMessage) a.getReplayExpectedMessages().peek()).pId);
         } else {
@@ -131,7 +135,6 @@ public class Actor implements Activity {
 
         if (a.firstMessage != null) {
           printMsg(a.firstMessage);
-
           if (a.mailboxExtension != null) {
             for (EventualMessage em : a.mailboxExtension) {
               printMsg(em);
@@ -150,6 +153,7 @@ public class Actor implements Activity {
         VM.println(a.getName() + " [" + a.getReplayActorId() + "] has " + n + " unexpected messages");
       }
     }
+    return result;
   }
 
   private static void printMsg(final EventualMessage msg) {
@@ -614,42 +618,6 @@ public class Actor implements Activity {
   @Override
   public String toString() {
     return "Actor";
-  }
-
-  public static final class DebugActor extends Actor {
-    // TODO: remove this tracking, the new one should be more efficient
-    private static final ArrayList<Actor> actors = new ArrayList<Actor>();
-    private final boolean isMain;
-    private final int id;
-
-    public DebugActor() {
-      super();
-      isMain = false;
-      synchronized (actors) {
-        actors.add(this);
-        id = actors.size() - 1;
-      }
-    }
-
-    @Override
-    protected void logMessageAddedToMailbox(final EventualMessage msg) {
-      VM.errorPrintln(toString() + ": queued task " + msg.toString());
-    }
-
-    @Override
-    protected void logMessageBeingExecuted(final EventualMessage msg) {
-      VM.errorPrintln(toString() + ": execute task " + msg.toString());
-    }
-
-    @Override
-    protected void logNoTaskForActor() {
-      VM.errorPrintln(toString() + ": no task");
-    }
-
-    @Override
-    public String toString() {
-      return "Actor[" + (isMain ? "main" : id) + "]";
-    }
   }
 
   public static class TracingActor extends Actor {
--- a/src/som/interpreter/actors/Actor.java
+++ b/src/som/interpreter/actors/Actor.java
@@ -385,6 +385,7 @@ public class Actor implements Activity {
           currentThread.currentMessageId += 1;
         }
 
+        currentThread.createdMessages += todo.size();
         ActorExecutionTrace.mailboxExecutedReplay(todo, baseMessageId, currentMailboxNo, actor);
     }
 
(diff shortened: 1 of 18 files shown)


smarr:
in the PR description, could you please clarify "some messages are processed out of order or not at all"? This sounds strange. Shouldn't all messages be processed in a deterministic order, predefined by the trace?

And, could you be explicit about what data is added to the trace? Would be good to track that in detail.

src/som/interpreter/actors/Actor.java
```diff
@@ -99,6 +114,57 @@ protected Actor() {
...
+  /**
+   * Prints a list of expected Messages and remaining mailbox content.
+   * @return true if there are actors expecting messages, false otherwise.
+   */
+  public static boolean printMissingMessages() {
+    if (!(VmSettings.REPLAY && VmSettings.DEBUG_MODE)) {
+      return false;
+    }
+
+    boolean result = false;
```
smarr: Could this could be better put into the `TracingActor`?

src/som/interpreter/actors/Actor.java
```diff
@@ -160,12 +226,39 @@ public synchronized void send(final EventualMessage msg) {
 
     logMessageAddedToMailbox(msg);
 
-    if (!isExecuting) {
+    // actor remains dormant until the expected message arrives
+    if ((!isExecuting) && replayCanProcess(msg)) {
```
smarr: could we solve this better by overriding `send()` in the `TracingActor`?
I'd like if all this trace and replay code would be in that class.

src/som/interpreter/actors/Actor.java
```diff
@@ -182,6 +275,11 @@ protected void logMessageAddedToMailbox(final EventualMessage msg) { }
   protected void logMessageBeingExecuted(final EventualMessage msg) { }
   protected void logNoTaskForActor() { }
   public long getActorId() { return 0; }
+  public long getReplayActorId() { return 0; }
```
smarr: Ideally, all these methods are only on the `TracingActor`.
If we can move most or all code related to it into that class, we might not need any of this in the basic actor class.

src/som/interpreter/actors/Actor.java
```diff
@@ -225,50 +328,111 @@ public void run() {
...
+        }
+
+        baseMessageId = currentThread.generateMessageBaseId(todo.size());
+        currentThread.currentMessageId = baseMessageId;
+
+        for (EventualMessage msg : todo) {
+          actor.logMessageBeingExecuted(msg);
+          currentThread.currentMessage = msg;
+
+          if (VmSettings.TRUFFLE_DEBUGGER_ENABLED) {
```
smarr: This breakpoint handling looks like duplicated code, could this be in a helper method?

src/som/interpreter/actors/Actor.java
smarr: Now I see the `ReplayActor`, or that one of course. Doesn't have to be the `TracingActor`, but, ideally only code that's really needed for all types of actors should be in the main class.

src/tools/concurrency/ActorExecutionTrace.java
```diff
@@ -42,13 +43,15 @@
...
 
   private static final int BUFFER_POOL_SIZE = Runtime.getRuntime().availableProcessors() * 4;
-  private static final int BUFFER_SIZE = 4096 * 1024;
+  static final int BUFFER_SIZE = 4096 * 1024;
   private static final byte MESSAGE_BASE = (byte) 0x80;
   private static final byte PROMISE_BIT = 0x40;
   private static final byte TIMESTAMP_BIT = 0x20;
   private static final byte PARAMETER_BIT = 0x10;
-  private static final int MESSAGE_SIZE = 45;
+  private static final int MESSAGE_SIZE = 45 + 16;
```
smarr: What's the size based on? Where do the magic numbers come from?
Would be good to have that clear in the code, so that it is obvious when one needs to change this number.

src/tools/concurrency/ActorExecutionTrace.java
```diff
@@ -153,15 +156,15 @@ public static synchronized void returnBuffer(final ByteBuffer b) {
...
+    Mailbox((byte) 5, 21),
 
     // at the beginning of buffer, allows to track what was created/executed
     // on which thread, really cheap solution, timestamp?
     Thread((byte) 6, 9),
 
     // for memory events another buffer is needed
     // (the gc callback is on Thread[Service Thread,9,system])
-    MailboxContd((byte) 7, 19),
+    MailboxContd((byte) 7, 23),
```
smarr: are those magic numbers related to the constants at the top of the file?

src/tools/concurrency/TraceParser.java
```diff
@@ -0,0 +1,342 @@
...
+public final class TraceParser {
+  private static final byte ACTOR_CREATION = 1;
+  private static final byte PROMISE_CREATION = 2;
+  private static final byte PROMISE_RESOLUTION = 3;
+  private static final byte PROMISE_CHAINED = 4;
+  private static final byte MAILBOX = 5;
+  private static final byte THREAD = 6;
+  private static final byte MAILBOX_CONTD = 7;
+
+  private static final byte MESSAGE_BIT = (byte) 0x80;
```
smarr: Why are these constants duplicated here? Can't we just put such things into a global place like `TraceData`? I suppose those constants need to remain in-sync across the whole program.

src/tools/concurrency/TraceParser.java
```diff
@@ -0,0 +1,342 @@
...
+    assert parser.mappedActors.containsKey(parentId) : "Parent doesn't exist";
+    return parser.mappedActors.get(parentId).getChild(childNo).actorId;
+  }
+
+  private TraceParser() {
+    assert VmSettings.REPLAY;
+  }
+
+  private void parseTrace() {
+    File f = new File(VmSettings.TRACE_FILE + ".trace");
```
smarr: could those two names please be a bit more explicit? Don't like such 1 or 2 letter names for important things.

src/tools/concurrency/TraceParser.java
```diff
@@ -0,0 +1,342 @@
...
+    File sf = new File(VmSettings.TRACE_FILE + ".sym");
+    HashMap<Long, List<Long>> unmappedActors = new HashMap<>(); // maps message to created actors
+    HashMap<Long, Queue<Long>> unmappedPromises = new HashMap<>(); // maps message to created promises
+
+    VM.println("Parsing Trace ...");
+
+    // create mapping from old to new symbol ids
+    try (FileInputStream sfos = new FileInputStream(sf);
```
smarr: sfos? what does this mean?

src/tools/concurrency/TraceParser.java
```diff
@@ -0,0 +1,342 @@
...
+        symbolMapping.put(Short.parseShort(a[0]), Symbols.symbolFor(a[1]));
+      });
+    } catch (FileNotFoundException e) {
+      throw new RuntimeException(e);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+
+    // parsing trace-file
+    try (FileInputStream fos = new FileInputStream(f); FileChannel channel = fos.getChannel()) {
```
smarr: fos? file output stream? but it's an input stream? what does fos mean?

src/tools/concurrency/TraceParser.java
```diff
@@ -0,0 +1,342 @@
...
+      channel.read(b);
+      b.flip();
+      while (channel.position() < channel.size() || b.remaining() > 0) {
+        if (!b.hasRemaining()) {
+          b.clear();
```
smarr: there are so many different things going in here. It's unclear what `b` is.

src/tools/concurrency/TraceParser.java
```diff
@@ -0,0 +1,342 @@
...
+      return children.get(childNo);
+    }
+
+    @Override
+    public String toString() {
+      return "" + actorId + ":" + childNo;
+    }
+  }
+
+  public static class Message implements Comparable<Message> {
```
smarr: Could we use another name than a plain 'Message' here?
Don't know, perhaps MessageRecord?

src/tools/concurrency/ActorExecutionTrace.java
daumayr: those number are used to determine if the buffer has enough free space for the event

Issue #86, opened 2017-01-12
Add support for Object Literals


smarr:
Object Literals are defined in the [Newspeak spec](http://bracha.org/newspeak-spec.pdf) sec. 5.1.10.

Depending on the context where they are defined, they either have the current activation (i.e. Truffle frame) as context, or `nil`. From that I derive that object literals can only be defined either on the module level or within a method.

The proposed syntax is `objectLiteral = (identifier, keywordMsg opt) opt, classBody.`

This means, we could have the following code:

```SOMns
public foo = (
  | local obj val |
  local := 1.
  obj := ( | public field = local. |
               local println.
         )(
           public class Baz = ()()
           public println = ( field println. local := local + 1. )
         ).

  obj println. (* => 1 *)
  local println. (* => 2 *)

 val := Value ()() (* => Signals a NotAValue exception *)
)
```

For the implementation, beside the parser changes, there is at least one change needed:
[`SClass` is defined](https://github.com/smarr/SOMns/blob/master/src/som/vmobjects/SClass.java#L65) with final SObjectWithClass enclosingObject to represent the outer scope. This needs to be changed so that the outer scope also can be a `MaterializedFrame`. This should be a straightforward change, there are only two things that depend on it: [reading of the outer object](https://github.com/smarr/SOMns/blob/master/src/som/interpreter/nodes/OuterObjectRead.java) which is either done with the `outer` keyword for an outer send, or for an implicit send. This does likely not need any adaptation, beside potential casts, because, access to local variables, i.e., the sends to the current activation are not done as 'sends' but are directly done as local variable accesses. I am not entirely clear on whether there are any issues in the parser's [lookup logic](https://github.com/smarr/SOMns/blob/master/src/som/compiler/MethodBuilder.java#L433) for determining whether it is a variable read or a message send, though. Currently, it first checks for local variables, and then does a send. That might be more involved with the additional nesting logic.

The second thing that needs the enclosing object is the [check whether a class is a value](https://github.com/smarr/SOMns/blob/master/src/som/primitives/ObjectPrims.java#L193). That might be a bit more involved. Currently it should be simply returning false for activations, I presume. Except perhaps if they don't have locals. That gets more complicated if we actually start supporting lazy slots (promise slots, not sure what the spec term is) and immutable slots, but, we currently don't have that.

I am not sure what else interferes with it, there might be more.

smarr:
This task is likely going to be worked on by [richard-roberts](https://github.com/) for Grace support.

[...]

PR #112, opened 2017-03-08
Implement object literals

richard-roberts:
Support for object literals (refer to #86). 

Object literals can now be declared with the `objL` keyword inside of:

- methods, `foo = (objL ()())` and
- block literals: `[objL ()()]`. 

Inheritance expressions can be used: `objL SuperName ()()` and statement separators are required when the declaration is not the last statement in a sequence: `objL ()(). objL ()()`.

The use of the keyword `objL` is intended as a temporary solution to avoid requiring more advanced look ahead in the parser (in particular, it is difficult to distinguish object literals from class expressions). Upon arriving at the `objL` token, the parser builds a `MixinDefinition` using `Parser.inheritanceListAndOrBody` method and provides this to an `ObjectLiteralNode`. This new literal node handles the instantiation of the object and provides the current frame as that literals context.

The logic for lookup of variables and message sends has been changed to account for the existence of object literals, as per the Newspeak specification. For any message send, the local variables are arguments of the current activation are first checked. If nothing is found, then the message is checked against the object literal's slots (via the new `hasSlotDefined()` method). Finally, if again nothing is found, then the local variables and arguments of all enclosing activations are checked.


src/som/compiler/MixinBuilder.java
```diff
@@ -165,16 +175,20 @@ public MixinScope getInstanceScope() {
...
 
-  public MixinBuilder getOuterBuilder() {
-    return outerBuilder;
+  public MixinBuilder getOuterMixinBuilder() {
+    return outerMixinBuilder;
+  }
+
+  public MethodBuilder getOuterMethodBuilder() {
+    return outerMethodBuilder;
   }
```
smarr: I wonder, shouldn't an outer builder be either a mixin or a method? It can't have both at the same time, or is that also useful?

src/som/interpreter/nodes/nary/RedirectReadNode.java
```diff
@@ -0,0 +1,29 @@
...
+import com.oracle.truffle.api.frame.VirtualFrame;
+import com.oracle.truffle.api.source.SourceSection;
+
+import som.compiler.Variable;
+import som.interpreter.nodes.PreevaluatedExpression;
+
+public class RedirectReadNode extends ExprWithTagsNode
+    implements PreevaluatedExpression {
+
+  final VirtualFrame frame;
```
smarr: First rule of Graal/Truffle: you can never ever save a reference to a virtual frame.
The `VirtualFrame` is supposed to never escape, because it is never supposed to be allocated as an object.
Furthermore, this is an AST node here, an AST node is normally independent of a specific activation.
Please see [`determineContext()`] (https://github.com/smarr/SOMns/blob/master/src/som/interpreter/nodes/ContextualNode.java#L51) and its uses to look up the relevant lexical frame for a variable in the lexical context at run time.

src/som/interpreter/nodes/nary/RedirectWriteNode.java
```diff
@@ -0,0 +1,49 @@
...
+import com.oracle.truffle.api.source.SourceSection;
+
+import som.compiler.Variable;
+import som.interpreter.nodes.ExpressionNode;
+import som.interpreter.nodes.PreevaluatedExpression;
+
+public class RedirectWriteNode extends ExprWithTagsNode
+    implements PreevaluatedExpression {
+
+  final VirtualFrame writeFrame;
```
smarr: Doesn't work (as for reads)

src/som/interpreter/nodes/nary/RedirectWriteNode.java
```diff
@@ -0,0 +1,49 @@
...
+  final ExpressionNode[] args;
+
+  public RedirectWriteNode(final VirtualFrame frame, final Variable var,
+      final ExpressionNode[] args, final SourceSection source) {
+    super(source);
+    this.writeFrame = frame;
+    this.args = args;
+
+    FrameSlot slotForVar = null;
+    for (FrameSlot slot : writeFrame.getFrameDescriptor().getSlots()) {
```
smarr: why is this different from the read case? why do you need lookup here?

src/som/interpreter/nodes/nary/RedirectWriteNode.java
```diff
@@ -0,0 +1,49 @@
...
+
+public class RedirectWriteNode extends ExprWithTagsNode
+    implements PreevaluatedExpression {
+
+  final VirtualFrame writeFrame;
+  final FrameSlot slot;
+  final ExpressionNode[] args;
+
+  public RedirectWriteNode(final VirtualFrame frame, final Variable var,
+      final ExpressionNode[] args, final SourceSection source) {
```
smarr: why is this an array?

src/som/vmobjects/SClass.java
```diff
@@ -56,6 +60,7 @@
   @CompilationFinal private HashSet<SlotDefinition> slots; // includes slots of super classes and mixins
 
   @CompilationFinal private MixinDefinition mixinDef;
+  @CompilationFinal private VirtualFrame context; // a frame containing the outer scope (implicit classes only)
```
smarr: Please check the [`context`](https://github.com/smarr/SOMns/blob/master/src/som/vmobjects/SBlock.java#L38) field in blocks (and the nodes that create block objects).
This frame needs to be *materialized*.

smarr: Kind of the same question as with the builders, why do you need a `context` and `enclosingObject`? Seems, we can only have one of those at the same time?

smarr:
Something that is likely to help working with anonymous classes: https://github.com/MetaConc/SOMns/commit/cbc907dc763cfc29f0794293889fc36da89d8f88

For blocks, I am now using the name of the outer methods to name blocks more precisely.

src/som/compiler/MixinBuilder.java
richard-roberts: I can't think of any reasonable case where both may be required. As an extreme example, I thought of one case (I believe it's nonsense, but for the sake of thinking about the problem):

```
class A  = ( |x = 1.| )(
  public foo = (
   | y |
  y := 11.
    
    ^ objL()( public printXY = (x print y println.) )
  )
)

class B = ( | x  = 2.| )(
  public bar = (
    | y |
   y := 12.
     (A foo) printXY.
  )
)
```

What should be printed when bar is invoked? If the object literal closes over both the method and the outer class, then bar will print (1 and 11). If it closes or neither then it will print (2 and 12). If it closes over only the method then, in theory, it could print 2 (B.x) and 11 (A.foo.x). 

If the correct answer is some reason 2 and 11, then we might need to store a reference to both the enclosing method and the outer class (so that during lookup we can check the current activation first, then bubble out to the activation of the enclosing method, then back to the enclosing classes: bar->foo->B instead of bar->foo->A). 

Again, I think this example might just be nonsensical, so there is likely no need to ever have both.

src/som/interpreter/nodes/nary/RedirectWriteNode.java
richard-roberts: The variable class contains behavior to read the value of variable from a frame, which I used to [evaluate and return ](https://github.com/richard-roberts/SOMns/blob/implement/objectLiterals/src/som/interpreter/nodes/nary/RedirectReadNode.java#L22) the value for the RedirectReadNode. There is no such behavior define to write a value to the frame (that I could find). The VirtualFrame implements a "setObject" method, which takes a  FrameSlot and a value. I need the lookup to find the FrameSlot that correlates to the variable I have.

Should I instead add behavior to the that handles writing in a similar way to [reading](https://github.com/richard-roberts/SOMns/blob/implement/objectLiterals/src/som/compiler/Variable.java#L142)?

src/som/compiler/MixinBuilder.java
smarr: Newspeak has only lexical scoping, not dynamic scoping.
Since you didn't declare any relation between `A` and `B`, the literal object does only know about `A`. But, in principle, it would close over all other outer scopes as well, if there would be any.

```
class OuterA = (|outX = 0.|) (
  class A  = ( |x = 1.| )(
    public foo = (
      | y |
      y := 11.
      ^ objL()(
          public printXY = (
            outX print. ' ' print.
            x print. ' ' print.
            y println.
      ) )
    )
  )
)

OuterA new A new foo printXY  ==>> '0 1 11\n'
```

src/som/interpreter/nodes/nary/RedirectWriteNode.java
smarr: the `Variable` class doesn't support writes, indeed. Not all variables allow assignment.
Only local variables are writable. Most importantly for you, Arguments are not assignable.

Here the relevant code in `Local`:
https://github.com/richard-roberts/SOMns/blob/implement/objectLiterals/src/som/compiler/Variable.java#L185

core-lib/TestSuite/LiteralObjectTests.som
```diff
@@ -0,0 +1,102 @@
...
+      | a b |
+      a := objL()().
+      b := objL()().
+      self deny: a == b.
+    )
+  
+    public testMethodLocalRead = (
+      | x a |
+      x := 1.
+      a := objL()(public b = (^ x)).
```
smarr: could we have space between parentheses with content?
`( public b = ( ^ x ) )`

core-lib/TestSuite/LiteralObjectTests.som
```diff
@@ -0,0 +1,102 @@
...
+      self assert: a b = 1.
```
smarr: I would prefer the use of `assert: a b equals: 1` makes for nicer error messages.
and I think, I'd drop the `self` since it isn't strictly necessary.

core-lib/TestSuite/LiteralObjectTests.som
```diff
@@ -0,0 +1,102 @@
...
+      x := 1.
+      y := [ objL()(public b = (x := x + 1. ^ x)) b ] value.
+      self assert: y = 2.
+    )
+
+    public testOuterKeywordInObjectLiteral = (
+      | x y |
+      x := objL()(public b = (^ outer ObjectTests)) b.
+      y := outer ObjectTests.
+      self assert: x == y.
```
smarr: better `assert: x is: y`

core-lib/TestSuite/LiteralObjectTests.som
```diff
@@ -0,0 +1,102 @@
...
+    public testMethodLocalWriteWhenNested = (
+      | x y |
+      x := 1.
+      y := [ objL()(public b = (x := x + 1. ^ x)) b ] value.
+      self assert: y = 2.
+    )
+
+    public testOuterKeywordInObjectLiteral = (
+      | x y |
+      x := objL()(public b = (^ outer ObjectTests)) b.
```
smarr: would be good to have a test without the outer keyword, too.
Just doing a message send to the outer object, implicitly, and seeing that it does the right thing.

core-lib/TestSuite/LiteralObjectTests.som
```diff
@@ -0,0 +1,102 @@
...
+      self assert: x == y.
+    )
+    
+    public testOuterKeywordInObjectLiteralInBlock = (
+      | x y |
+      x := [objL()(public b = (^ outer ObjectTests)) b] value.
+      y := outer ObjectTests.
+      self assert: x == y.
+    )
+
```
smarr: how about a test assigning to the field on the other object?
Is there a test that explicitly tests that the order in the lookup is correct? especially for multiple nested objects, that we get the object field/variable for the scope?

src/som/compiler/MixinBuilder.java
```diff
@@ -385,10 +382,11 @@ public MixinDefinition assemble(final SourceSection source) {
...
 
   private boolean outerScopeIsImmutable() {
-    if (outerBuilder == null) {
+    if (getFirstOuterMixin() == null) {
```
smarr: should that be `isModule()`? that would an unrelated change improvement, but I noticed, because there you used `getContext()` instead of `getFirstOuterMixin()`.

src/som/compiler/Parser.java
```diff
@@ -1444,6 +1448,38 @@ private LiteralNode literalString() throws ParseError {
...
 
+  /**
+   * Generates an implicit class nested onto the outer class.
+   *
+   * @return an expression that instantiates the implicit class
+   */
+  private ExpressionNode literalObject(final MethodBuilder builder)
+      throws ProgramDefinitionError {
+
+    // TODO: make note regarding the specification.
```
smarr: Please do :)

src/som/compiler/Parser.java
```diff
@@ -1444,6 +1448,38 @@ private LiteralNode literalString() throws ParseError {
     return new StringLiteralNode(s, getSource(coord));
   }
 
+  /**
+   * Generates an implicit class nested onto the outer class.
```
smarr: nested onto? sounds strange. Also, that's not correct anymore, is it?

src/som/compiler/Parser.java
```diff
@@ -1444,6 +1448,38 @@ private LiteralNode literalString() throws ParseError {
...
+    MethodBuilder primaryFactory = implicitClassBuilder.getPrimaryFactoryMethodBuilder();
+    primaryFactory.addArgument("self", getEmptySource());
+    primaryFactory.setSignature(Symbols.NEW);
+    implicitClassBuilder.setupInitializerBasedOnPrimaryFactory(source);
+    inheritanceListAndOrBody(implicitClassBuilder);
+
+    // Generate the object literal node.
+    MixinDefinition mixinDef = implicitClassBuilder.assemble(source);
+    ObjectLiteralNode objectLiteral = new ObjectLiteralNode(mixinDef, builder.getSelfRead(source), source);
+    return SNodeFactory.createMessageSend(symbolFor("new"),
```
smarr: hm, the structure is unintuitive to me. I think, I'd move the `#new` into the object literal node. So, the result of the object literal node is an object, and not a class.

src/som/compiler/Parser.java
```diff
@@ -1482,7 +1518,10 @@ private ExpressionNode nestedBlock(final MethodBuilder builder) throws ProgramDe
       blockPattern(builder);
     }
 
-    String outerMethodName = stripColons(builder.getOuterBuilder().
+    if (builder.getFirstOuterMethod() == null) {
+      throw new java.lang.NullPointerException("The outer context of a block should never be null");
```
smarr: perhaps this can be turned into an assertion?
```Java
assert builder.getFirstOuterMethod() == null: "The outer context of a block should never be null";
```

core-lib/TestSuite/LiteralObjectTests.som
```diff
@@ -0,0 +1,102 @@
+(* Copyright (c) 2001-2015 see AUTHORS file
```
smarr: also, you might want to add yourself to the AUTHORS file while you're at it.
And, the CHANGELOG.md deserves an entry, too :)

smarr:
I tried to benchmark this in the `LanguageFeatures.som` file.

The first two are revealing bugs.
The third shows missing splitting support.

```Smalltalk
  public class ObjectLiterals = Benchmark ()(
    public benchmark = (
      | cnt |
      cnt := 0.
      1 to: 1000 do: [:i |
        objL()( public inc = ( cnt := cnt + 1 ) ) inc.
      ].
      ^ cnt
    )

    public verifyResult: result = (
      ^ 1000 = result
    )
  ) : (
    public newInstance = ( ^ self new )
    public setupVerifiedRun: run = ( run innerIterations: 1 )
  )

  public class ObjectLiterals2 = Benchmark (
  | private cnt ::= 0. |
  )(
    private inc = (
      objL()( public inc = ( cnt := cnt + 1 ) ) inc.
    )

    public benchmark = (
      cnt := 0.
      1 to: 1000 do: [:i | inc ].
      ^ cnt
    )

    public verifyResult: result = (
      ^ 1000 = result
    )
  ) : (
    public newInstance = ( ^ self new )
    public setupVerifiedRun: run = ( run innerIterations: 1 )
  )

  public class ObjectLiterals3 = Benchmark ()(
    private inc: i = (
      | cnt |
      cnt := i.
      objL()( public inc = ( cnt := cnt + 1 ) ) inc.
      ^ cnt
    )

    public benchmark = (
      | cnt |
      cnt := 0.
      1 to: 1000 do: [:i | cnt := inc: cnt ].
      ^ cnt
    )

    public verifyResult: result = (
      ^ 1000 = result
    )
  ) : (
    public newInstance = ( ^ self new )
    public setupVerifiedRun: run = ( run innerIterations: 1 )
  )
```

core-lib/TestSuite/LiteralObjectTests.som
smarr: We'd also want to test giving a super class to the literal, and that the primary factory method initialize it correctly.

src/som/compiler/MixinBuilder.java
richard-roberts: Yes, it should be.

core-lib/Hello.som
```diff
@@ -1,8 +1,27 @@
-class Hello usingPlatform: platform = Value ()(
+class Hello usingPlatform: platform = Value (
+)(
   public main: args = (
-    'Hello World!' println.
-    args from: 2 to: args size do: [ :arg | arg print. ' ' print ].
-    '' println.
-    ^ 0
+    | x2 |
```
smarr: Please don't change the hello world :)
Feel free to add more tests though, also possibly 'basic interpreter tests' if that makes debugging easier (see the folder and JUnit tests).

core-lib/TestSuite/LiteralObjectTests.som
smarr: Comment still stands.

core-lib/TestSuite/LiteralObjectTests.som
```diff
@@ -0,0 +1,112 @@
...
+THE SOFTWARE.
+*)
+class ObjectTests usingPlatform: platform testFramework: minitest = (
+| private TestContext = minitest TestContext. |
+) (
+  public class ObjectTest = TestContext ()(
+
+    public testEquality = (
+      | a b |
+      a := objL()().
```
smarr: I have been thinking about the syntax.
You are currently not supporting super classes for object literals, if I see it correctly, right?

How about making the syntax then simply `obj(`?
This is not a valid sequence of symbols.
I don't think there is any legal expression that first has a unary message send or something, and then a parenthetical expression.
So, we could simply lex `obj(` as `StartObjectLiteral` or something like that?

core-lib/TestSuite/LiteralObjectTests.som
```diff
@@ -0,0 +1,102 @@
...
+      b := objL()().
```
smarr: TODO?

core-lib/TestSuite/LiteralObjectTests.som
smarr: !

core-lib/TestSuite/LiteralObjectTests.som
smarr: !

core-lib/TestSuite/LiteralObjectTests.som
smarr: !

src/som/compiler/MethodBuilder.java
```diff
@@ -381,6 +381,22 @@ private int getContextLevel(final String varName) {
     throw new IllegalStateException("Didn't find variable.");
   }
 
+  private int getThruContextLevel(final String varName) {
```
smarr: What does this mean? `getThruContextLevel`? Don't understand.

src/som/compiler/MethodBuilder.java
```diff
@@ -408,6 +424,35 @@ protected Variable getVariable(final String varName) {
     return null;
   }
 
+  protected Variable getThruVariable(final String varName) {
```
smarr: Don't understand what that is supposed to mean either.

src/som/compiler/MethodBuilder.java
```diff
@@ -455,17 +515,26 @@ public ExpressionNode getImplicitReceiverSend(final SSymbol selector,
...
       return SNodeFactory.createMessageSend(selector,
           new ExpressionNode[] {getSelfRead(source)}, false, source, null, language);
-    } else {
-      // otherwise, it is an implicit receiver send
-      return SNodeFactory.createImplicitReceiverSend(selector,
-          new ExpressionNode[] {getSelfRead(source)},
-          getCurrentMethodScope(), getEnclosingMixinBuilder().getMixinId(),
-          source, language.getVM());
     }
+
```
smarr: One empty line only, please. Same below.

src/som/compiler/MethodBuilder.java
```diff
@@ -514,6 +600,31 @@ protected Local getLocal(final String varName) {
...
+      return locals.get(varName);
+    }
+
+    if (outerBuilder != null) {
+      Local outerLocal = outerBuilder.getLocal(varName);
+      if (outerLocal != null) {
+        accessesVariablesOfOuterScope = true;
+        return outerLocal;
+      }
+
```
smarr: No empty line please

src/som/compiler/MixinDefinition.java
```diff
@@ -714,4 +716,38 @@ public Object instantiateObject(final Object... args) {
   public String toString() {
     return name.getString() + "[" + sourceSection + "]";
   }
+
+  public void cloneAndAdaptAfterScopeChange(final InliningVisitor inliner) {
+
```
smarr: where is the cloning? you modify the scope before cloning? doesn't seem right

src/som/interpreter/InliningVisitor.java
```diff
@@ -73,6 +73,10 @@ public MethodScope getScope(final Method method) {
     return scope.getScope(method);
   }
 
+  public MethodScope getScopeDirectly() {
```
smarr: Naming unclear

src/som/interpreter/LexicalScope.java
```diff
@@ -267,8 +301,7 @@ public void setMethod(final Method method) {
       CompilerAsserts.neverPartOfCompilation("LexicalContext.sOM()");
       // might be reset when doing inlining/embedded, but should always
       // refer to the same method
-      assert this.method == null ||
-          this.method.getSourceSection() == method.getSourceSection();
+      //assert this.method == null || this.method.getSourceSection() == method.getSourceSection();
```
smarr: Neither condition is valid anymore?
Why not remove the line then?
Though, i think the second might hold?

src/som/interpreter/Method.java
```diff
@@ -51,7 +51,7 @@ public Method(final String name, final SourceSection sourceSection,
     this.definition = definition;
     this.block = block;
     this.methodScope = methodScope;
-    assert methodScope.isFinalized();
+    //assert methodScope.isFinalized();
```
smarr: if this doesn't hold anymore, tell me why, and remove the assertion, no commented out code in this repo!!!!!!

src/som/interpreter/nodes/ContextualNode.java
```diff
@@ -48,17 +50,35 @@ public final int getContextLevel() {
...
   @ExplodeLoop
-  protected final MaterializedFrame determineContext(final VirtualFrame frame) {
-    SBlock self = (SBlock) SArguments.rcvr(frame);
+  protected final MaterializedFrame determineContext(final Frame frame) {
+    Object next = SArguments.rcvr(frame);
     int i = contextLevel - 1;
 
     while (i > 0) {
-      self = (SBlock) self.getOuterSelf();
+      if (next instanceof SBlock) {
```
smarr: Graal is going to hate this code. Did you ran the benchmarks to see what's going to happen?

src/som/interpreter/nodes/literals/ObjectLiteralNode.java
```diff
@@ -0,0 +1,56 @@
...
+import som.vmobjects.SAbstractObject;
+import som.vmobjects.SBlock;
+import som.vmobjects.SClass;
+import som.vmobjects.SObjectWithClass;
+
+
+public class ObjectLiteralNode extends LiteralNode {
+
+  private final MixinDefinition mixinDef;
+  @Child private ExpressionNode selfRead;
```
smarr: Please order fields so that child nodes and other fields form separate groups.

src/som/compiler/Variable.java
```diff
@@ -57,6 +57,9 @@ public boolean equals(final Object o) {
       assert name.equals(var.name) : "Defined in the same place, but names not equal?";
       return true;
     }
+    if (var.name == "self" && name == "self") {
```
smarr: So, you are saying that the source information for `self` nodes aren't correct?
I'd try to fix those, instead of adding the hack here (which is also not correct/reliable Java as the codacy bot indicates).
It could well be that there are issues with `self` since there are nodes constructed somewhat manually in various places, which is something not happening with other variables.

src/som/interpreter/nodes/literals/BlockNode.java
```diff
@@ -66,7 +66,7 @@ public SInvokable getBlockMethod() {
 
   @Override
   public SBlock executeSBlock(final VirtualFrame frame) {
-    return new SBlock(blockMethod, null, blockClass);
+    return new SBlock(blockMethod, frame.materialize(), blockClass);
```
smarr: The whole point of this class is to not require the frame.
So, this is wrong here.
Instead, you should make sure that [`requiresContext()`](https://github.com/smarr/SOMns/blob/dev/src/som/compiler/MethodBuilder.java#L223) returns `true`, when the frame is needed

src/som/interpreter/nodes/literals/ObjectLiteralNode.java
```diff
@@ -30,7 +32,20 @@ public ObjectLiteralNode(final MixinDefinition mixinDef, final ExpressionNode ne
...
+     * itself (in cases where the receiver is a class) or the `outerSelf`
+     * (in cases where the receiver is a block).
+     */
+    SAbstractObject rcvr = (SAbstractObject) SArguments.rcvr(frame);
+    SObjectWithClass outer;
+    if (rcvr instanceof SObjectWithClass) {
+      outer = (SObjectWithClass) rcvr;
+    } else if (rcvr instanceof SBlock) {
+      outer = (SObjectWithClass) ((SBlock) rcvr).getOuterSelf();
+    } else {
```
smarr: why don't you just use a node here that does the right thing?
There should a node to read outer, no need to duplicate that logic here, or is it somehow special?

src/som/compiler/MixinDefinition.java
```diff
@@ -715,4 +716,59 @@ public Object instantiateObject(final Object... args) {
   public String toString() {
     return name.getString() + "[" + sourceSection + "]";
   }
+
+  public MixinDefinition cloneAndAdaptAfterScopeChange(final MethodScope adaptedScope, final int appliesTo) {
+
+    MixinScope adaptedInstanceScope = new MixinScope(instanceScope.getOuterMixin(), adaptedScope);
+    MixinScope adaptedClassScope = new MixinScope(classScope.getOuterMixin(), adaptedScope);
+    adaptedInstanceScope.setMixinDefinition(this, false);
```
smarr: This looks broken.
Shouldn't you set the clone here on the scopes?

src/som/compiler/MixinDefinition.java
```diff
@@ -715,4 +716,59 @@ public Object instantiateObject(final Object... args) {
...
+        nestedMixinDefinitions, mixinId, accessModifier,
+        adaptedInstanceScope, adaptedClassScope,
+        allSlotsAreImmutable, outerScopeIsImmutable,
+        isModule, sourceSection);
+
+    adaptedInstanceScope.setMixinDefinition(clone, false);
+    adaptedClassScope.setMixinDefinition(clone, true);
+
+      for (SSymbol key : clone.factoryMethods.keySet()) {
+        SInvokable method = clone.factoryMethods.get(key);
```
smarr: this screams for a helper method.

src/som/compiler/MixinDefinition.java
```diff
@@ -715,4 +716,59 @@ public Object instantiateObject(final Object... args) {
...
+
+      if (!(d instanceof SInvokable)) {
+        continue;
+      }
+
+      SInvokable method = (SInvokable) d;
+      Method blockIvk = (Method) method.getInvokable();
+      MethodScope m = blockIvk.getLexicalScope().split(adaptedScope);
+
+      Method adapted = blockIvk.cloneAndAdaptAfterScopeChange(
```
smarr: Is this the same code as above in the loop?

src/som/compiler/MethodBuilder.java
richard-roberts: I removed the extra line, unsure why this comment has not become outdated.

core-lib/TestSuite/LiteralObjectTests.som
```diff
@@ -0,0 +1,197 @@
...
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*)
+class ObjectTests usingPlatform: platform testFramework: minitest = (
+| private TestContext = minitest TestContext.
+  private outerVariable ::= 1. |
+) (
+  
+  public class FakeSuperClass = (
```
smarr: Why is this called `FakeSuperClass`? It seems to be used as a normal super class for the literal, no?
So, what's fake about it?

core-lib/TestSuite/LiteralObjectTests.som
```diff
@@ -0,0 +1,197 @@
...
+
+THE SOFTWARE IS PROVIDED 'AS IS', WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*)
+class ObjectTests usingPlatform: platform testFramework: minitest = (
```
smarr: Literal objects or object literals? We should decide on one and stick with it.
File is name LiteralObjectTests. The module name is ObjectTests.
I think, I'd use ObjectLiteralTests for both.

core-lib/TestSuite/LiteralObjectTests.som
```diff
@@ -0,0 +1,197 @@
...
+| private TestContext = minitest TestContext.
+  private outerVariable ::= 1. |
+) (
+  
+  public class FakeSuperClass = (
+    | variable ::= 0. |
+    variable:: 1.
+  )(
+    public getVariable = (
```
smarr: Any specific reason why you wanted an extra method here, instead of just making the slot public?

src/som/interpreter/nodes/literals/ObjectLiteralNode.java
```diff
@@ -0,0 +1,52 @@
...
+    this.mixinDef = mixinDef;
+    this.outerRead = outerRead;
+    this.newMessage = (AbstractMessageSendNode) newMessage;
+    this.superClassResolver = mixinDef.getSuperclassAndMixinResolutionInvokable().createCallTarget();
+  }
+
+  @Override
+  public Object executeGeneric(final VirtualFrame frame) {
+    SObjectWithClass outer = (SObjectWithClass) outerRead.executeGeneric(frame);
+    Object superclassAndMixins = superClassResolver.call(outer);
```
smarr: This should probably be cached, similar to normal object instantiation.
Was there a specific reason why you didn't reuse the node that's doing the instantiation?

src/som/vmobjects/SClass.java
```diff
@@ -63,14 +64,29 @@
   @CompilationFinal private ClassFactory instanceClassGroup; // the factory for this object
 
   protected final SObjectWithClass enclosingObject;
+  private MaterializedFrame context;
```
smarr: This should be final, no? Why can't you set the context during construction?

src/som/vmobjects/SClass.java
```diff
@@ -63,14 +64,29 @@
...
+
+  public void setContext(final MaterializedFrame context) {
+    this.context = context;
+  }
+
+  public boolean hasContext() {
+    return context != null;
+  }
+
+  public MaterializedFrame recallContext() {
```
smarr: `recallContext`? why not `getContext()`?

src/som/vmobjects/SObjectWithContext.java
```diff
@@ -0,0 +1,16 @@
...
+import com.oracle.truffle.api.frame.MaterializedFrame;
+
+/**
+ * Something that has a context.
+ * Used by SBlock and SClass currently.
+ */
+
+public interface SObjectWithContext {
+
+  MaterializedFrame getContext();
```
smarr: Please specify the semantics for both methods.

We need to get better with documentation.

core-lib/TestSuite/LiteralObjectTests.som
richard-roberts: Hah, it's indeed a real super class - my mistake.

src/som/vmobjects/SClass.java
richard-roberts: So it's different from the ObjectWithContext's `getContext()`. The subclass `SClass` is the class that receives the frame, but the superclass `SObjectWithClass` implements the `ObjectWithContext` method.

src/som/interpreter/nodes/literals/ObjectLiteralNode.java
richard-roberts: I was unsure if it was possible for the outer object of a literal to change, I'm guessing not so I will make these cache when this node is first called.

src/som/vmobjects/SClass.java
richard-roberts: Yes, it should be final. I've added a method to the `ClassInstantationNode` to enable this.

smarr:
Hm, just curious, it is on purpose that your commits are attributed with two different email addresses?

src/som/interpreter/nodes/literals/ObjectLiteralNode.java
```diff
@@ -0,0 +1,77 @@
...
+  /**
+   * This method caches the outer object and superclass and mixins used to instantiate the object
+   * literal. Note that this method is only called once when this node is invoked for the first
+   * time.
+   */
+  private void cacheOuterAndSuperClass(final VirtualFrame frame) {
+    assert outer == null && superclassAndMixins == null :
+      "The outer object and superclass and mixins can only be cached once.";
+
+    outer = (SObjectWithClass) outerRead.executeGeneric(frame);
```
smarr: Ehm, I think we misunderstood us here.
This seems like caching that's going to break semantics.

You can't instantiate the same object literal multiple times now correctly, even when the outer scope has changed, you're binding it to the first outer object you see.

src/som/interpreter/nodes/literals/ObjectLiteralNode.java
```diff
@@ -0,0 +1,77 @@
...
+    superclassAndMixins = superClassResolver.call(outer);
```
smarr: And this is also not safe.

Please see `ClassInstantiationNode` again, there are guards to cache things.
Perhaps you can adapt `ClassInstantiationNode` to support your case so you don't need to roll your own?

smarr:
Ok, so with the last changes, performance seems to be ok again. There might be minor regressions, but it is unclear from the numbers whether it is noise or something in need to be fixed.

If this is merged, there should be a new issue created that works out some remaining issues.
I'd really like if the scope handling is completely refactored and redone.
There should be only a single linear chain of scopes, this PR however introduces complications for object literals.
And, once that is worked out, the contextual nodes should get the chain of casts to lookup the context created by the parser. All knowledge is there, but because of the complications with the scopes, it's a little tricky to get it right for inlining at the moment.

PR #113, opened 2017-03-08
Introduce TracingActivityThread base class

smarr:
Move thread state relevant for tracing of different kinds of activities (actors, processes,…) into common base class.

Use TracingActivityThread where possible, especially in ActorExecutionTrace.

- rely on `assert` instead of `if` to test for thread
- ignore `java.util.TimeThread` for recording promises, doesn't seem to be necessary

@daumayr please have a brief look at this, that's a change I originally introduced in my huge Kompos PR #106 and now extracted, because I had issues getting replay to work again. Seems to work now.


daumayr:
I just had a look at it, seems ok

smarr:
thanks

PR #118, opened 2017-03-13
Added Breaking/Erroring of Promises

salenaer:
I've added ruining of promises, currently, an error occurs in the tests, but not during regular use. Neither Carmen nor I find the reason of the error.
Can you have a look at the code and see if we missed something?

### Design: Ruining Promise in Absence of Handlers

A message is sent from A to B. The evaluation of this message in B results in an error. As such the promise (the result of this message send in A) needs to be ruined. To ruin a promise means setting the state to erroneous and running all error handlers connected to this promise. This exposes the occurrence of the error to A.

First, the handlers bound to this promise are started. If no such handlers exist the chained promises (promises resolved with the current ruined promise) are ruined. If none of the chained promises handled the exception, this exception is uncaught. Normally the user should be informed when an uncaught exception occurs. However, in our case, this is impossible, because of scheduling it is possible that the promise is ruined before the handlers are added. This is a race condition, this race condition should not be exposed to the user. 

Pseudocode example of race condition as a result of scheduling, time runs downwards

```
A 						B
promise := B <-: message

						process(message)


promise whenResolved: [something] onError: [ handlers]
```

To change this behaviour make ruinChainedPromise return a handled boolean. Based on this boolean perform desired behaviour inside onError method of SPromise.


src/som/interpreter/actors/ReceivedRootNode.java
```diff
@@ -47,6 +52,22 @@ protected final void resolvePromise(final VirtualFrame frame,
     resolve.executeEvaluated(frame, resolver, result, isBreakpointOnPromiseResolution);
   }
 
+  protected final void ruinPromise(final VirtualFrame frame,
+      final SResolver resolver, final SAbstractObject exception) {
+    // lazy initialization of resolution node
+    if (ruin == null) {
+      CompilerDirectives.transferToInterpreterAndInvalidate();
+      if (resolver == null) {
+        throw new SomException(exception); //I'm ruining a promise without a resolver. No onError can be added so rethrow the exception
```
smarr: This is wrong behavior, this will crash the thread.
If the user decided not to capture the result, we can't crash the VM. This is part of the language semantics. A `NullResolver` should handle this case, like in the normal resolution case.

src/som/interpreter/actors/ReceivedRootNode.java
```diff
@@ -47,6 +52,22 @@ protected final void resolvePromise(final VirtualFrame frame,
     resolve.executeEvaluated(frame, resolver, result, isBreakpointOnPromiseResolution);
   }
 
+  protected final void ruinPromise(final VirtualFrame frame,
```
smarr: I am still not ok with `ruin` as the term here.
Please use `break` instead.

src/som/interpreter/actors/ResolvePromiseNode.java
```diff
@@ -50,8 +51,8 @@ public SResolver chainedPromise(final VirtualFrame frame, final SResolver resolv
         normalResolution(resolver, promiseValue.getValueUnsync(),
            isBreakpointOnPromiseResolution);
       } else if (promiseValue.isErroredUnsync()) {
-        CompilerDirectives.transferToInterpreter();
-        throw new NotYetImplementedException();
+          CompilerDirectives.transferToInterpreter();
```
smarr: Why do you change the whitespace here.
Indentation is 2 spaces, everywhere.

src/som/interpreter/actors/ResolvePromiseNode.java
```diff
@@ -82,7 +83,7 @@ public SResolver normalResolution(final SResolver resolver, final Object result,
 
   public static void resolve(final WrapReferenceNode wrapper,
       final SPromise promise, final Object result, final Actor current,
-      final boolean isBreakpointOnPromiseResolution) {
+      final boolean isBreakpointOnPromiseResolution) {
```
smarr: Please enable whitespace trimming in your editor.

src/som/interpreter/actors/SPromise.java
```diff
@@ -22,7 +23,7 @@
 
 public class SPromise extends SObjectWithClass {
   private enum Resolution {
-    UNRESOLVED, ERRORNOUS, SUCCESSFUL, CHAINED
+    UNRESOLVED, ERRONEOUS, SUCCESSFUL, CHAINED
```
smarr: Thank you

src/som/interpreter/actors/SPromise.java
```diff
@@ -139,8 +140,8 @@ private void registerMoreWhenResolved(final PromiseMessage msg) {
 
   public synchronized void registerOnError(final PromiseMessage msg,
       final Actor current) {
-    if (resolutionState == Resolution.ERRORNOUS) {
-      scheduleCallbacksOnResolution(value, msg, current, false);
+    if (resolutionState == Resolution.ERRONEOUS) {
+        scheduleCallbacksOnResolution(value, msg, current, false);
```
smarr: Whitespace.

src/som/interpreter/actors/SPromise.java
```diff
@@ -328,8 +329,74 @@ public static void setSOMClass(final SClass cls) {
       resolverClass = cls;
     }
 
-    public final void onError() {
-      throw new NotYetImplementedException(); // TODO: implement
+    public static void onError(final SAbstractObject exception, final Object wrapped, final SPromise p, final Actor current) {
+      //System.out.println("called on error");
```
smarr: Please don't commit commented debugging code.

src/som/interpreter/actors/SPromise.java
```diff
@@ -328,8 +329,74 @@ public static void setSOMClass(final SClass cls) {
...
+          p.value = exception;
+          p.resolutionState = Resolution.ERRONEOUS;
+        }
+
+        boolean handled = false;
+        
+        //execute all exceptionHandlers if the type of wrapped is equal to the exceptionClass the handler handles.
+        if (p.onException != null) {
+          for (int i = 0; i < p.onException.size(); i++) {
+            SClass exceptionClass = p.onException.get(i);
```
smarr: I was under the impression that we figured out that the `onException` stuff can't be supported because it implies visible race conditions. Or is this not the case after all?
I guess it depends on how this is realized. But from the sequential world, filtering would be the way to go, which is a race. So, it can't be done, I think.

I'd rather remove the `on:` exception support.

src/som/interpreter/actors/SPromise.java
```diff
@@ -328,8 +329,74 @@ public static void setSOMClass(final SClass cls) {
...
+          for (int i = 0; i < p.onError.size(); i++) {
+            handled = true;
+            PromiseMessage handle = p.onError.get(i);
+            p.scheduleCallbacksOnResolution(exception, handle, current, false);
+          }
+        }
+        return handled;
+    }
+
+    protected static void ruinChainedPromisesUnsynch(final SAbstractObject exception, final SPromise promise, final Actor current){
```
smarr: Would it be possible to use the existing code here instead, and pass as an argument whether the resolution is successful or erroneous? This seems to be a duplication of rather intricate code, which I'd really like to avoid.

src/som/primitives/actors/PromisePrims.java
```diff
@@ -182,6 +182,26 @@ public final SPromise whenResolvedOnError(final SPromise promise,
       return whenResolvedOrError(promise, resolved, error, resolvedTarget,
           errorTarget, registerNode);
     }
+
```
smarr: Please don't included dead code

thesis/behaviour.som
```diff
@@ -0,0 +1,45 @@
+class Behaviour usingPlatform: platform = Value (
```
smarr: Are the following files really to be committed here? seems like something you'd like in a separate repo.
Considering that the folder is named `thesis`?

salenaer: I want them in my local fork and thought you could refuse them when merging with the main.

core-lib/TestSuite/ActorTests.som
```diff
@@ -596,4 +609,223 @@ class ActorTests usingPlatform: platform testFramework: minitest = Value (
...
+      wrPromise := promisePair promise whenResolved: [ :v | NoExceptionReceivedException signal . ] onError: [ :ex | ex . ]. 
+
+      assertPromise := wrPromise whenResolved: [:v | assert: v msg equals: exception msg . ] .
+
+      promisePair ruin: exception.
+      ^ assertPromise
+    ) *)
+
+  (*    
+    public testAsyncPromiseOfPromiseResolvesToException = (
```
salenaer: Most of the tests are commented out as they do work when ran on their own. When I run multiple at the same time I get some error I can't seem to solve.

thesis/behaviour.som
smarr: Ok, you might want to put them in a separate repo to avoid issues, since you seem to have sent the PR from master, I am worried that if i delete them, you might just miss them.

core-lib/TestSuite/ActorTests.som
smarr: Ok, so, just to clarify, do you want me to change the things I currently marked as issues and also try to solve the remaining bugs? I could do that. But I don't want to just take it out of your hands, since you probably spend already a lot of sweat on the PR.

salenaer: You can take over, I really need to focus on things I need for my thesis

smarr:
@salenaer thanks again for the pull request.
And apologies for many of the issues that you saw. Some of that is because of broken stuff in the test framework.  For instance: https://github.com/smarr/SOMns/pull/118/commits/92afbcd198d4e9209c63d8db7d6126c6bd4e395b

As you may see from the updated PR, I tried to avoid much of the code duplication of the previous version by reusing all of the normal resolution code. Instead of duplicating it, it now got a parameter that defines success or failure. Beside that, I only replicated a few optimizations that were in place before.

After a few minor fixes to the tests, they pass.

Would be great, if you could review the PR and check whether I missed anything.

Thanks!

smarr:
And yeah, that's just another example for: everything that's not tested is broken. Sorry...

src/som/primitives/actors/PromisePrims.java
```diff
@@ -121,84 +120,83 @@ protected static final SPromise whenResolved(final SPromise rcvr,
...
   @Primitive(primitive = "actorsWhen:resolved:onError:")
   public abstract static class WhenResolvedOnErrorPrim extends TernaryExpressionNode {
-    @Child protected RegisterWhenResolved registerNode = new RegisterWhenResolved();
+    @Child protected RegisterWhenResolved registerWhenResolved = new RegisterWhenResolved();
+    @Child protected RegisterOnError registerOnError = new RegisterOnError();
 
     public WhenResolvedOnErrorPrim(final boolean eagWrap, final SourceSection source) { super(eagWrap, source); }
 
-    @Specialization(guards = {"resolvedMethod == resolved.getMethod()", "errorMethod == error.getMethod()"})
+    @Specialization(guards = {"resolvedMethod == resolved.getMethod()", "errorMethod == error.getMethod()"}, limit = "1000")
```
smarr: Todo for myself: remove limit, and replace by proper fallback case

smarr:
@salenaer the PR is merged and on master now. Thanks.

PR #119, opened 2017-03-13
Add button to unfold code pane 

smarr:
Allow to unfold the code pane for activities that have not hit a breakpoint yet.
Thus, enable exploration of source code for activity even if not stopped.

Think, the best would be to identify the allocation/start site of the activity, and show this site.

### DIFF
--- a/src/tools/concurrency/ActorExecutionTrace.java
+++ b/src/tools/concurrency/ActorExecutionTrace.java
@@ -26,6 +26,7 @@ import javax.management.NotificationListener;
 import javax.management.openmbean.CompositeData;
 
 import com.oracle.truffle.api.CompilerDirectives.TruffleBoundary;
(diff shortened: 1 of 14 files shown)


src/tools/concurrency/ActorExecutionTrace.java
```diff
@@ -210,6 +211,7 @@ public static final void forceSwapBuffers() {
 
   protected enum Events {
     ActorCreation(TraceData.ACTOR_CREATION,         19),
+    ActorCreationWithOrigin(TraceData.ACTOR_CREATION_ORIGIN,         27),
```
smarr: Just for change documentation: I think, I'll change the encoding here, and write after the normal actor creation event, directly following, a creation origin. That avoids having everything three times.

PR #121, opened 2017-03-15
Add Kompos tests for broken promises and breakpoints

smarr:
- [ ] add Kompos tests for promise breakpoints
- [ ] add tests for the various new breakpoint types for non-actor models


smarr:
Was done with: https://github.com/smarr/SOMns/pull/69

PR #125, opened 2017-03-23
Select activity colors by type

smarr:
Chose from the lighter colors.

New look is something like this:
screen shot 2017-03-23 at 18 37 25


smarr:
Merged as https://github.com/smarr/SOMns/commit/e05be43eff4a6c6b464903c05e48091f16f924b0

PR #129, opened 2017-04-10
Message stepping

ctrlpz:
This pull request adds support for message stepping.
Specifically were added four operations: 

1. step-to-message-receiver: can be trigger only from the EventualSendNode. Stops execution in the receiver of the message.
2. step-to-promise-resolution: can be trigger from any expression that creates promises. Stops execution at promise resolution, before executing the callback. 
3. step-return-from-turn-to-promise-resolution: can be trigger from any StatementTag inside a turn. Stops execution at promise resolution after processing the turn.
4. step-to-next-turn: can be trigger from anywhere inside a turn. Stops execution in the next message of the mailbox of the current actor.

The main changes are:

- add in `SteppingType` class the implementation corresponding to the four stepping operations for actors.
- field `promiseRegisteredOn` in `PromiseCallbackMessage` class for support the step-return-from-turn-to-promise-resolution operation.
- in `EventualSendNode` class the method `hasMessageReceiverBreakpoint` is called when the messages are created, to check if was triggered any stepping operation and it updates the corresponding breakpoint flags. Similarly, in `PromisePrims` class a method `hasPromiseResolutionBreakpoint` is called to check the stepping and updates the corresponding breakpoint flags.
- add in `SteppingStrategy` class the method `isEnabled` to check if a given strategy has been enabled.
- in the `TracingActor` class is the method for handling breakpoints and stepping. This method was modified in order to check the flag of the actor corresponding to the step-to-next-turn and also to check if a step-return-from-turn-to-promise-resolution was executed.

TODO
- implement stepping tests.

Notes: 
- an improvement of the stepping visualization is going to be implemented in another branch.
- the message conditional breakpoint, and step until an operation is going to be implemented in another branch.

### DIFF
--- a/src/som/interpreter/actors/Actor.java
+++ b/src/som/interpreter/actors/Actor.java
@@ -187,12 +187,6 @@ public class Actor implements Activity {
     mailboxExtension.append(msg);
   }
 
-  protected static void handleBreakPoints(final EventualMessage msg, final WebDebugger dbg) {
-    if (VmSettings.TRUFFLE_DEBUGGER_ENABLED && msg.isMessageReceiverBreakpointSet()) {
-      dbg.prepareSteppingUntilNextRootNode();
-    }
-  }
-
   /**
    * Is scheduled on the fork/join pool and executes messages for a specific
    * actor.
@@ -268,7 +262,9 @@ public class Actor implements Activity {
     private void execute(final EventualMessage msg,
         final ActorProcessingThread currentThread, final WebDebugger dbg, final int i) {
       currentThread.currentMessage = msg;
-      handleBreakPoints(msg, dbg);
+      if (VmSettings.ACTOR_TRACING) {
+        TracingActor.handleBreakpointsAndStepping(msg, dbg, actor);
+      }
 
       if (i >= 0 && VmSettings.MESSAGE_TIMESTAMPS) {
         executionTimeStamps[i] = System.currentTimeMillis();
@@ -325,6 +321,8 @@ public class Actor implements Activity {
     }
   }
 
+  @Override
+  public void setStepToNextTurn(final boolean val) {  }
 
   public static final class ActorProcessingThreadFactory implements ForkJoinWorkerThreadFactory {
     @Override
--- a/src/som/interpreter/actors/EventualMessage.java
+++ b/src/som/interpreter/actors/EventualMessage.java
@@ -9,14 +9,11 @@ import som.VM;
 import som.interpreter.actors.Actor.ActorProcessingThread;
 import som.interpreter.actors.ReceivedMessage.ReceivedCallback;
 import som.interpreter.actors.SPromise.SResolver;
-import som.vm.Symbols;
 import som.vmobjects.SBlock;
 import som.vmobjects.SSymbol;
 
 
 public abstract class EventualMessage {
-  private static final SSymbol VALUE_SELECTOR = Symbols.symbolFor("value:");
-
   protected final Object[]  args;
   protected final SResolver resolver;
   protected final RootCallTarget onReceive;
@@ -233,13 +230,18 @@ public abstract class EventualMessage {
   /** The callback message to be send after a promise is resolved. */
   public static final class PromiseCallbackMessage extends PromiseMessage {
     protected final SPromise promise;
+    /**
+     * The promise on which this callback is registered on.
+     */
+    protected final SPromise promiseRegisteredOn;
 
     public PromiseCallbackMessage(final long causalMessageId, final Actor owner, final SBlock callback,
         final SResolver resolver, final RootCallTarget onReceive, final boolean triggerMessageReceiverBreakpoint,
-        final boolean triggerPromiseResolverBreakpoint, final SPromise parent) {
+        final boolean triggerPromiseResolverBreakpoint, final SPromise parent, final SPromise promiseRegisteredOn) {
       super(causalMessageId, new Object[] {callback, null}, owner, resolver, onReceive,
           triggerMessageReceiverBreakpoint, triggerPromiseResolverBreakpoint);
       this.promise = parent;
+      this.promiseRegisteredOn = promiseRegisteredOn;
     }
 
     @Override
@@ -276,6 +278,10 @@ public abstract class EventualMessage {
     @Override
     public SPromise getPromise() {
       return promise;
+    }
+
+    public SPromise getPromiseRegisteredOn() {
+      return promiseRegisteredOn;
     }
   }
 
--- a/src/som/interpreter/actors/EventualSendNode.java
+++ b/src/som/interpreter/actors/EventualSendNode.java
@@ -10,6 +10,7 @@ import com.oracle.truffle.api.dsl.Cached;
 import com.oracle.truffle.api.dsl.Specialization;
 import com.oracle.truffle.api.frame.VirtualFrame;
 import com.oracle.truffle.api.instrumentation.Instrumentable;
+import com.oracle.truffle.api.instrumentation.StandardTags.StatementTag;
 import com.oracle.truffle.api.nodes.ExplodeLoop;
 import com.oracle.truffle.api.nodes.Node;
 import com.oracle.truffle.api.source.SourceSection;
@@ -32,6 +33,10 @@ import som.vm.constants.Nil;
 import som.vmobjects.SSymbol;
 import tools.concurrency.Tags.EventualMessageSend;
 import tools.concurrency.Tags.ExpressionBreakpoint;
+import tools.debugger.SteppingStrategy;
+import tools.debugger.SteppingStrategy.ReturnFromTurnToPromiseResolution;
+import tools.debugger.SteppingStrategy.ToMessageReceiver;
+import tools.debugger.SteppingStrategy.ToPromiseResolution;
 import tools.debugger.nodes.AbstractBreakpointNode;
 import tools.debugger.session.Breakpoints;
 
@@ -186,7 +191,8 @@ public class EventualSendNode extends ExprWithTagsNode {
       DirectMessage msg = new DirectMessage(
           EventualMessage.getCurrentExecutingMessageId(), target, selector, args,
           owner, resolver, onReceive,
-          messageReceiverBreakpoint.executeCheckIsSetAndEnabled(), promiseResolverBreakpoint.executeCheckIsSetAndEnabled());
+          hasMessageReceiverBreakpoint(resolver), promiseResolverBreakpoint.executeCheckIsSetAndEnabled());
+
       target.send(msg, actorPool);
     }
 
@@ -197,8 +203,34 @@ public class EventualSendNode extends ExprWithTagsNode {
       PromiseSendMessage msg = new PromiseSendMessage(
           EventualMessage.getCurrentExecutingMessageId(), selector, args,
           rcvr.getOwner(), resolver, onReceive,
-          messageReceiverBreakpoint.executeCheckIsSetAndEnabled(), promiseResolverBreakpoint.executeCheckIsSetAndEnabled());
+          hasMessageReceiverBreakpoint(resolver), promiseResolverBreakpoint.executeCheckIsSetAndEnabled());
+
       registerNode.register(rcvr, msg, rcvr.getOwner());
+    }
+
+    /**
+     * Check if any stepping strategy has been set and updates the corresponding breakpoint flag.
+     * If the strategy ToMessageReceiver is active, the flag messageReceiverBreakpoint is updated.
+     * If the strategy ToPromiseResolution is active, the flag promiseResolutionBreakpoint is updated.
+     * If the strategy ReturnFromTurnToPromiseResolution is active, the flag triggerStopBeforeExecuteCallback
+     * of the promise of the causal message is updated.
+     * Returns the flag of the message receiver breakpoint.
+     */
+    private boolean hasMessageReceiverBreakpoint(final SResolver resolver) {
+      boolean stepReceiver = SteppingStrategy.isEnabled(ToMessageReceiver.class);
+      boolean stepResolution = SteppingStrategy.isEnabled(ToPromiseResolution.class);
+      boolean stepReturnFromTurn = SteppingStrategy.isEnabled(ReturnFromTurnToPromiseResolution.class);
+
+      boolean msgRcvrBkp = messageReceiverBreakpoint.executeCheckIsSetAndEnabled() || stepReceiver;
+      if (resolver != null && !resolver.getPromise().isTriggerPromiseResolutionBreakpoint() && stepResolution) {
+        resolver.getPromise().setTriggerPromiseResolutionBreakpoint(stepResolution);
+      }
+      EventualMessage causalMsg = EventualMessage.getCurrentExecutingMessage();
+      if (causalMsg.getResolver() != null && stepReturnFromTurn) {
+        causalMsg.getResolver().getPromise().setTriggerStopBeforeExecuteCallback(stepReturnFromTurn);
+      }
+
+      return msgRcvrBkp;
     }
 
     protected RegisterWhenResolved createRegisterNode() {
@@ -226,7 +258,6 @@ public class EventualSendNode extends ExprWithTagsNode {
     public final SPromise toPromiseWithResultPromise(final Object[] args,
         @Cached("createRegisterNode()") final RegisterWhenResolved registerNode) {
       SPromise rcvr = (SPromise) args[0];
-
       SPromise  promise  = SPromise.createPromise(EventualMessage.getActorCurrentMessageIsExecutionOn(), promiseResolutionBreakpoint.executeCheckIsSetAndEnabled(), false, false);
       SResolver resolver = SPromise.createResolver(promise);
 
@@ -244,7 +275,8 @@ public class EventualSendNode extends ExprWithTagsNode {
       DirectMessage msg = new DirectMessage(EventualMessage.getCurrentExecutingMessageId(),
           current, selector, args, current,
           resolver, onReceive,
-          messageReceiverBreakpoint.executeCheckIsSetAndEnabled(), promiseResolverBreakpoint.executeCheckIsSetAndEnabled());
+          hasMessageReceiverBreakpoint(resolver), promiseResolverBreakpoint.executeCheckIsSetAndEnabled());
+
       current.send(msg, actorPool);
 
       return result;
@@ -273,14 +305,15 @@ public class EventualSendNode extends ExprWithTagsNode {
       DirectMessage msg = new DirectMessage(EventualMessage.getCurrentExecutingMessageId(),
           current, selector, args, current,
           null, onReceive,
-          messageReceiverBreakpoint.executeCheckIsSetAndEnabled(), promiseResolverBreakpoint.executeCheckIsSetAndEnabled());
+          hasMessageReceiverBreakpoint(null), promiseResolverBreakpoint.executeCheckIsSetAndEnabled());
+
       current.send(msg, actorPool);
       return Nil.nilObject;
     }
 
     @Override
     protected boolean isTaggedWith(final Class<?> tag) {
-      if (tag == EventualMessageSend.class || tag == ExpressionBreakpoint.class) {
+      if (tag == EventualMessageSend.class || tag == ExpressionBreakpoint.class || tag == StatementTag.class) {
         return true;
       }
       return super.isTaggedWith(tag);
--- a/src/som/primitives/actors/PromisePrims.java
+++ b/src/som/primitives/actors/PromisePrims.java
@@ -8,6 +8,7 @@ import com.oracle.truffle.api.dsl.GenerateNodeFactory;
 import com.oracle.truffle.api.dsl.ImportStatic;
 import com.oracle.truffle.api.dsl.NodeFactory;
 import com.oracle.truffle.api.dsl.Specialization;
+import com.oracle.truffle.api.instrumentation.StandardTags.StatementTag;
 import com.oracle.truffle.api.nodes.DirectCallNode;
 import com.oracle.truffle.api.source.SourceSection;
 
@@ -79,7 +80,7 @@ public final class PromisePrims {
     public final SImmutableObject createPromisePair(final Object nil,
         @Cached("create()") final DirectCallNode factory) {
 
-      SPromise promise   = SPromise.createPromise(EventualMessage.getActorCurrentMessageIsExecutionOn(), promiseResolutionBreakpoint.executeCheckIsSetAndEnabled(), promiseResolverBreakpoint.executeCheckIsSetAndEnabled(), true);
+      SPromise promise   = SPromise.createPromise(EventualMessage.getActorCurrentMessageIsExecutionOn(), SPromise.hasPromiseResolutionBreakpoint(promiseResolutionBreakpoint), promiseResolverBreakpoint.executeCheckIsSetAndEnabled(), true);
       SResolver resolver = SPromise.createResolver(promise);
       return (SImmutableObject) factory.call(new Object[] {SPromise.pairClass, promise, resolver});
     }
@@ -88,11 +89,12 @@ public final class PromisePrims {
 
     @Override
     protected boolean isTaggedWithIgnoringEagerness(final Class<?> tag) {
-      if (tag == CreatePromisePair.class || tag == ExpressionBreakpoint.class) {
+      if (tag == CreatePromisePair.class || tag == ExpressionBreakpoint.class || tag == StatementTag.class) {
         return true;
       }
       return super.isTaggedWithIgnoringEagerness(tag);
     }
+
   }
 
   // TODO: can we find another solution for megamorphics callers that
@@ -142,11 +144,11 @@ public final class PromisePrims {
 
       Actor current = EventualMessage.getActorCurrentMessageIsExecutionOn();
 
-      SPromise  promise  = SPromise.createPromise(current, promiseResolutionBreakpoint.executeCheckIsSetAndEnabled(), false, false);
+      SPromise  promise  = SPromise.createPromise(current, SPromise.hasPromiseResolutionBreakpoint(promiseResolutionBreakpoint), false, false);
       SResolver resolver = SPromise.createResolver(promise);
 
       PromiseCallbackMessage pcm = new PromiseCallbackMessage(EventualMessage.getCurrentExecutingMessageId(), rcvr.getOwner(),
-          block, resolver, blockCallTarget, false, promiseResolverBreakpoint.executeCheckIsSetAndEnabled(), rcvr);
+          block, resolver, blockCallTarget, false, promiseResolverBreakpoint.executeCheckIsSetAndEnabled(), promise, rcvr);
       registerNode.register(rcvr, pcm, current);
 
       return promise;
@@ -154,7 +156,7 @@ public final class PromisePrims {
 
     @Override
     protected boolean isTaggedWithIgnoringEagerness(final Class<?> tag) {
-      if (tag == WhenResolved.class || tag == ExpressionBreakpoint.class) {
+      if (tag == WhenResolved.class || tag == ExpressionBreakpoint.class || tag == StatementTag.class) {
         return true;
       }
       return super.isTaggedWithIgnoringEagerness(tag);
@@ -198,11 +200,11 @@ public final class PromisePrims {
 
       Actor current = EventualMessage.getActorCurrentMessageIsExecutionOn();
 
-      SPromise  promise  = SPromise.createPromise(current, promiseResolutionBreakpoint.executeCheckIsSetAndEnabled(), false, false);
+      SPromise  promise  = SPromise.createPromise(current, SPromise.hasPromiseResolutionBreakpoint(promiseResolutionBreakpoint), false, false);
       SResolver resolver = SPromise.createResolver(promise);
 
       PromiseCallbackMessage msg = new PromiseCallbackMessage(EventualMessage.getCurrentExecutingMessageId(), rcvr.getOwner(),
-          block, resolver, blockCallTarget, false, promiseResolverBreakpoint.executeCheckIsSetAndEnabled(), promise);
+          block, resolver, blockCallTarget, false, promiseResolverBreakpoint.executeCheckIsSetAndEnabled(), promise, rcvr);
       registerNode.register(rcvr, msg, current);
 
       return promise;
@@ -210,7 +212,7 @@ public final class PromisePrims {
 
     @Override
     protected boolean isTaggedWithIgnoringEagerness(final Class<?> tag) {
-      if (tag == OnError.class || tag == ExpressionBreakpoint.class) {
+      if (tag == OnError.class || tag == ExpressionBreakpoint.class || tag == StatementTag.class) {
         return true;
       }
       return super.isTaggedWithIgnoringEagerness(tag);
@@ -263,11 +265,11 @@ public final class PromisePrims {
 
       Actor current = EventualMessage.getActorCurrentMessageIsExecutionOn();
 
-      SPromise  promise  = SPromise.createPromise(current, promiseResolutionBreakpoint.executeCheckIsSetAndEnabled(), false, false);
+      SPromise  promise  = SPromise.createPromise(current, SPromise.hasPromiseResolutionBreakpoint(promiseResolutionBreakpoint), false, false);
       SResolver resolver = SPromise.createResolver(promise);
 
-      PromiseCallbackMessage onResolved = new PromiseCallbackMessage(EventualMessage.getCurrentExecutingMessageId(), rcvr.getOwner(), resolved, resolver, resolverTarget, false, promiseResolverBreakpoint.executeCheckIsSetAndEnabled(), rcvr);
-      PromiseCallbackMessage onError    = new PromiseCallbackMessage(EventualMessage.getCurrentExecutingMessageId(), rcvr.getOwner(), error,    resolver, errorTarget,    false, promiseResolverBreakpoint.executeCheckIsSetAndEnabled(), rcvr);
+      PromiseCallbackMessage onResolved = new PromiseCallbackMessage(EventualMessage.getCurrentExecutingMessageId(), rcvr.getOwner(), resolved, resolver, resolverTarget, false, promiseResolverBreakpoint.executeCheckIsSetAndEnabled(), promise, rcvr);
+      PromiseCallbackMessage onError    = new PromiseCallbackMessage(EventualMessage.getCurrentExecutingMessageId(), rcvr.getOwner(), error,    resolver, errorTarget,    false, promiseResolverBreakpoint.executeCheckIsSetAndEnabled(),  promise, rcvr);
 
       synchronized (rcvr) {
         registerWhenResolved.register(rcvr, onResolved, current);
@@ -278,10 +280,11 @@ public final class PromisePrims {
 
     @Override
     protected boolean isTaggedWithIgnoringEagerness(final Class<?> tag) {
-      if (tag == WhenResolvedOnError.class || tag == ExpressionBreakpoint.class) {
+      if (tag == WhenResolvedOnError.class || tag == ExpressionBreakpoint.class || tag == StatementTag.class) {
         return true;
       }
       return super.isTaggedWithIgnoringEagerness(tag);
     }
+
   }
 }
--- a/src/tools/debugger/entities/SteppingType.java
+++ b/src/tools/debugger/entities/SteppingType.java
@@ -1,18 +1,25 @@
 package tools.debugger.entities;
 
 import com.google.gson.annotations.SerializedName;
+import com.oracle.truffle.api.instrumentation.StandardTags.StatementTag;
 
 import som.vm.NotYetImplementedException;
 import tools.concurrency.Tags;
 import tools.concurrency.Tags.ActivityCreation;
 import tools.concurrency.Tags.ChannelRead;
 import tools.concurrency.Tags.ChannelWrite;
+import tools.concurrency.Tags.EventualMessageSend;
+import tools.concurrency.Tags.ExpressionBreakpoint;
 import tools.debugger.SteppingStrategy.AfterCommit;
 import tools.debugger.SteppingStrategy.IntoSpawn;
 import tools.debugger.SteppingStrategy.ReturnFromActivity;
+import tools.debugger.SteppingStrategy.ReturnFromTurnToPromiseResolution;
 import tools.debugger.SteppingStrategy.ToChannelOpposite;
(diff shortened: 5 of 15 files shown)


src/som/interpreter/actors/Actor.java
```diff
@@ -182,11 +187,34 @@ protected void appendToMailbox(final EventualMessage msg) {
     mailboxExtension.append(msg);
   }
 
-  protected static void handleBreakPoints(final EventualMessage msg, final WebDebugger dbg) {
-    if (VmSettings.TRUFFLE_DEBUGGER_ENABLED && msg.isBreakpoint()) {
-      dbg.prepareSteppingUntilNextRootNode();
+  protected static void handleBreakpointsAndStepping(final EventualMessage msg, final WebDebugger dbg) {
+    if (VmSettings.TRUFFLE_DEBUGGER_ENABLED) {
+      // check step-into and breakpoint before the message is processed
+      if (msg.isStepInto()) {
```
smarr: if you feel that the comment is needed here, perhaps the method `isStepInto` can be renamed so that the comment isn't necessary anymore.

src/som/interpreter/actors/Actor.java
```diff
@@ -182,11 +187,34 @@ protected void appendToMailbox(final EventualMessage msg) {
...
+        dbg.prepareStepIntoMessage();
+      } else if (msg.isBreakpoint()) {
```
smarr: `isBreakpoint()` sounds now very unspecific, I think it would be good to rename to clarify what it actually checks.
Also, it is unclear to me that `isBreakpoint()` and `isStepInto()` two are truly mutually exclusive.

src/tools/debugger/session/Breakpoints.java
```diff
@@ -222,4 +236,59 @@ public static AbstractBreakpointNode createOpposite(final SourceSection source)
...
+    }
+
+    return null;
+  }
+
+  /**
+   * Disable existing breakpoints and runs execution before the next root node is executed.
+   */
+  public void prepareStepIntoMessage() {
+    disableBreakpoints();
```
smarr: why do you disable breakpoints? I can't imagine that this is what a user wants.
If I am stepping, and there are breakpoints sets, why wouldn't I want them to trigger?

src/som/interpreter/actors/Actor.java
```diff
@@ -182,11 +187,34 @@ protected void appendToMailbox(final EventualMessage msg) {
...
+        dbg.prepareSteppingUntilNextRootNode();
+      }
+      // check if a step-over has been made in the previous message
+      if (stopNextMessage) {
+       dbg.prepareStepIntoMessage();
```
smarr: after looking at the code, I think all three conditions, this one and above, just want a `dbg.prepareSteppingUntilNextRootNode();`

src/som/interpreter/actors/Actor.java
```diff
@@ -182,11 +187,34 @@ protected void appendToMailbox(final EventualMessage msg) {
...
+      }
+      // check step-return
+      if (msg.isStepReturn()) {
```
smarr: comment doesn't seem to be needed, the method name says it all, no?

src/som/interpreter/actors/Actor.java
```diff
@@ -182,11 +187,34 @@ protected void appendToMailbox(final EventualMessage msg) {
...
+          dbg.prepareStepReturnMessage(msg.getCausalMessage());
+      }
+      // check type of callback if a step-return has been made
+      if (msg instanceof PromiseSendMessage) {
+        if (((PromiseSendMessage) msg).getPromise().isTriggerStepReturnOnCallbacks()) {
```
smarr: instead of having a comment, I think it would be better to have a helper method that does the check, and where the name represents what it is supposed to do, i.e., encodes the essence of the comment.

src/som/interpreter/actors/Actor.java
```diff
@@ -182,11 +187,34 @@ protected void appendToMailbox(final EventualMessage msg) {
...
+          dbg.prepareSteppingUntilNextRootNode();
+        }
+      } else if (msg instanceof PromiseCallbackMessage) {
+        if (((PromiseCallbackMessage) msg).getOriginalPromise().isTriggerStepReturnOnCallbacks()) {
```
smarr: Same here, this should really be a helper method, possibly as part of the previous one.

src/som/interpreter/actors/Actor.java
```diff
@@ -182,11 +187,34 @@ protected void appendToMailbox(final EventualMessage msg) {
...
+          dbg.prepareSteppingUntilNextRootNode();
```
smarr: I don't think I want to see more than a single call to `prepareSteppingUntilNextRootNode()` in this method

src/som/interpreter/actors/Actor.java
```diff
@@ -241,11 +269,13 @@ protected void processCurrentMessages(final ActorProcessingThread currentThread,
 
       try {
         execute(firstMessage, currentThread, dbg, -1);
+        stopNextMessage = firstMessage.isStepOver();
```
smarr: this should only be done when debugging is enabled, no?
and since you do it also in the loop below, should it be part of the `execute()` method?

src/som/interpreter/actors/EventualMessage.java
```diff
@@ -20,6 +21,7 @@
   protected final Object[]  args;
   protected final SResolver resolver;
   protected final RootCallTarget onReceive;
+  protected final EventualMessage causalMessage;
   protected final long causalMessageId;
```
smarr: do we really need two fields here?
I suppose `causalMessageId` is part of the `causalMessage`?

src/som/interpreter/actors/EventualMessage.java
```diff
@@ -233,13 +245,15 @@ public SPromise getPromise() {
   /** The callback message to be send after a promise is resolved. */
   public static final class PromiseCallbackMessage extends PromiseMessage {
     protected final SPromise promise;
+    protected final SPromise originalPromise;
```
smarr: I don't know what `originalPromise` is suppose to mean.
How does it relate to the `promise`? These names need clarification, I think.

src/som/interpreter/actors/EventualMessage.java
```diff
@@ -345,4 +366,24 @@ public boolean isBreakpoint() {
...
+     return this.triggerStepping == SteppingType.STEP_RETURN;
+  }
+
+  public boolean isStepInto() {
+    return this.triggerStepping == SteppingType.STEP_INTO;
+ }
+
+  public SteppingType getTriggerStepping() {
+    return this.triggerStepping;
+  }
```
smarr: The `isStep*` methods and `getTriggerStepping()` seem to be somehow redundant.
Do we really need all variants. I'd think that we have either the `is*` methods, or `getTriggerStepping()`, but probably not both?

src/som/interpreter/actors/EventualSendNode.java
```diff
@@ -133,6 +134,23 @@ protected SendNode(final SSymbol selector, final WrapReferenceNode[] wrapArgs,
     }
 
     /**
+     * Update messageReceiverBreakpoint in case that a step into has been made.
+     */
+    private void checkSteppingInto() {
+      this.messageReceiverBreakpoint = insert(Breakpoints.createReceiver(this.source));
```
smarr: This node seems to be created every time there is a STEP_INTO, is this really necessary? Isn't it sufficient to create it once only?

src/som/interpreter/actors/EventualSendNode.java
```diff
@@ -175,21 +193,30 @@ protected void sendDirectMessage(final Object[] args, final Actor owner,
...
 
+      SteppingType step = getStepping();
+
       DirectMessage msg = new DirectMessage(
-          EventualMessage.getCurrentExecutingMessageId(), target, selector, args,
+          EventualMessage.getCurrentExecutingMessage(), EventualMessage.getCurrentExecutingMessageId(), target, selector, args,
           owner, resolver, onReceive,
           messageReceiverBreakpoint.executeCheckIsSetAndEnabled(), promiseResolverBreakpoint.executeCheckIsSetAndEnabled());
+
+      msg.setTriggerStepping(step);
```
smarr: Why do you set the `step` on the message after its creation, and don't pass it in the constructor?
Also, I guess the field in the message object could be final then?

src/som/interpreter/actors/EventualSendNode.java
```diff
@@ -175,21 +193,30 @@ protected void sendDirectMessage(final Object[] args, final Actor owner,
...
+      SteppingType step = getStepping();
+
       PromiseSendMessage msg = new PromiseSendMessage(
-          EventualMessage.getCurrentExecutingMessageId(), selector, args,
+          EventualMessage.getCurrentExecutingMessage(), EventualMessage.getCurrentExecutingMessageId(),
+          selector, args,
           rcvr.getOwner(), resolver, onReceive,
           messageReceiverBreakpoint.executeCheckIsSetAndEnabled(), promiseResolverBreakpoint.executeCheckIsSetAndEnabled());
+
+      msg.setTriggerStepping(step);
```
smarr: could be set via the constructor?

src/som/interpreter/actors/EventualSendNode.java
```diff
@@ -233,10 +260,14 @@ public final SPromise toNearRefWithResultPromise(final Object[] args) {
...
-      DirectMessage msg = new DirectMessage(EventualMessage.getCurrentExecutingMessageId(),
-          current, selector, args, current,
+      SteppingType step = getStepping();
+
+      DirectMessage msg = new DirectMessage(EventualMessage.getCurrentExecutingMessage(),
+          EventualMessage.getCurrentExecutingMessageId(), current, selector, args, current,
           resolver, onReceive,
           messageReceiverBreakpoint.executeCheckIsSetAndEnabled(), promiseResolverBreakpoint.executeCheckIsSetAndEnabled());
+
+      msg.setTriggerStepping(step);
```
smarr: could be set via the constructor?

src/som/interpreter/actors/EventualSendNode.java
```diff
@@ -262,10 +293,14 @@ public final Object toPromiseWithoutResultPromise(final Object[] args,
...
-      DirectMessage msg = new DirectMessage(EventualMessage.getCurrentExecutingMessageId(),
-          current, selector, args, current,
+      SteppingType step = getStepping();
+
+      DirectMessage msg = new DirectMessage(EventualMessage.getCurrentExecutingMessage(),
+          EventualMessage.getCurrentExecutingMessageId(), current, selector, args, current,
           null, onReceive,
           messageReceiverBreakpoint.executeCheckIsSetAndEnabled(), promiseResolverBreakpoint.executeCheckIsSetAndEnabled());
+
+      msg.setTriggerStepping(step);
```
smarr: could be set via the constructor?

src/tools/debugger/session/Breakpoints.java
```diff
@@ -161,8 +169,14 @@ public boolean evaluate() {
 
  public synchronized BreakpointEnabling<MessageReceiverBreakpoint> getReceiverBreakpoint(
       final FullSourceCoordinate section) {
-    return receiverBreakpoints.computeIfAbsent(section,
+   BreakpointEnabling<MessageReceiverBreakpoint> bkp = receiverBreakpoints.computeIfAbsent(section,
         ss -> new BreakpointEnabling<>(new MessageReceiverBreakpoint(false, section)));
+
+   if (bkp.isDisabled() && stepping != null) {
+     bkp.setEnabled(stepping.getSteppingTypeOperation(section) == SteppingType.STEP_INTO);
```
smarr: why is the enabling status of the breakpoint depending on some stepping info? and how can that stepping info be global? shouldn't it be local to an activity/actor?

src/tools/debugger/session/Breakpoints.java
```diff
@@ -54,13 +59,16 @@
...
+  private Stepping stepping;
+
   public Breakpoints(final Debugger debugger, final WebDebugger webDebugger) {
     this.truffleBreakpoints           = new HashMap<>();
     this.receiverBreakpoints          = new HashMap<>();
     this.promiseResolverBreakpoints   = new HashMap<>();
     this.promiseResolutionBreakpoints = new HashMap<>();
     this.channelOppositeBreakpoint    = new HashMap<>();
     this.debuggerSession = debugger.startSession(webDebugger);
+    this.stepping = null;
```
smarr: I don't think it is possible to have global stepping field. I don't see how that can be meaningful.

src/tools/debugger/session/Breakpoints.java
```diff
@@ -222,4 +236,59 @@ public static AbstractBreakpointNode createOpposite(final SourceSection source)
...
+    disableBreakpoints();
+    prepareSteppingUntilNextRootNode();
+  }
+
+  /**
+   * Disable existing breakpoints and set true flag in the promise of a message that was step return.
+   *
+   */
+  public void prepareStepReturnFromMessage(final EventualMessage msg) {
+    disableBreakpoints();
```
smarr: don't think this is a good idea.

src/tools/debugger/stepping/Stepping.java
```diff
@@ -0,0 +1,42 @@
...
+   * Return the stepping operation that has been made for a specific source section,
+   * based on the start line of the source section.
+   * It can be only compared the start line because in the case that a sequential stepping operation
+   * was made before a message stepping operation, the source section fields of column and char length
+   * does not correspond to the asynchronous message operator source section.
+   */
+  public SteppingType getSteppingTypeOperation(final FullSourceCoordinate source) {
+
+    if (stepActor != null && source.startLine == stepActor.getSourceSection().startLine) {
+      if (stepActor.getSteppingType() == SteppingType.STEP_INTO) {
```
smarr: This looks strange, why don't you just return the value of `getSteppingType()`?

src/tools/debugger/stepping/Stepping.java
```diff
@@ -0,0 +1,42 @@
...
+    STEP_RETURN
+  }
+
+  private StepActorOperation stepActor;
+
+  public Stepping(final StepActorOperation stepActor) {
+    this.stepActor = stepActor;
+  }
+  /**
+   * Return the stepping operation that has been made for a specific source section,
```
smarr: I don't understand this comment.

tools/kompos/src/ui-controller.ts
```diff
@@ -298,4 +298,31 @@ export class UiController extends Controller {
     this.view.onContinueExecution(act);
     this.vmConnection.sendDebuggerAction("return", act);
   }
+
+  public stepIntoMessage(actId: string) {
+    const activityId = getActivityIdFromView(actId);
```
smarr: In all these methods, everything is duplicated but one string. Please don't do this!

src/som/interpreter/actors/EventualMessage.java
ctrlpz: I put the clarification as a comment in getOriginalPromise() method: The promise on which this callback is registered on. I need it to check if a stepReturn was trigger for that promise.

src/tools/debugger/stepping/Stepping.java
ctrlpz: I tried to explain in the comment why the comparison is not made with the sourceSection object, but only with one of its attributes, the start line. I probably need to think more about the combination of both stepping sequential and message-oriented, but as I see it now I think I need this comparison in order to execute message stepping operations when the previous step was sequential. For example, if you did a step-over and after you want to do a step-into-message operation, the sourceSection of the SuspendedEvent does not match with the sourceSection of the <-:

src/tools/debugger/session/Breakpoints.java
ctrlpz: true, I thought it was a case that it was needed but I verified and it isn't.

src/som/interpreter/actors/EventualMessage.java
smarr: that doesn't help me when reading the code here. Please use better names.

src/tools/debugger/stepping/Stepping.java
smarr: This sounds wrong. How do you know that the next statement is on the next line? it could as well be a line later, or on the same line, no?

src/som/interpreter/actors/EventualMessage.java
ctrlpz: But `EventualMessage` does not have an `id` field, unless we use EventualMessage.getCurrentExecutingMessageId(), but I think that this method will return the `id` of the current executing message (as the name says), if we want the id of the causal message we should save it when is execute it, right? This is used in the TraceBuffer class.

smarr: ok. @daumayr can we move the `causalMessageId` into the `causalMessage`? That could also avoid that it is looked up multiple times via `EventualMessage.getCurrentExecutingMessageId()`? Might be a small performance win even?

src/tools/debugger/stepping/Stepping.java
ctrlpz: yes, is true...then I need indeed compare with the sourceSection...I will think more about this

smarr: hmmm, this all still looks strange. I am not really sure what you want to achieve with that method in the first place, you use that in the end to set the type of stepping strategy on the message? I am confused. sorry, won't be able to look into this in detail today.

ctrlpz: no, please don't look anymore, now I'm looking at the different scenarios with the breakpoints and I'm changing the implementation. I think the principal problem I had is that I did not consider the stepping for each actor/activity. Thanks for your comments, they have been helpful.

src/som/interpreter/actors/EventualMessage.java
daumayr: You can, but you would need a tracing specialization for all the messages

src/som/interpreter/actors/Actor.java
```diff
@@ -75,6 +77,9 @@ public static Actor createActor(final VM vm) {
   /** Is scheduled on the pool, and executes messages to this actor. */
   protected final ExecAllMessages executor;
 
+  /** Flag that indicates if a step-to-next-turn action has been made in the previous message. */
+  protected boolean stepToNextTurn;
```
smarr: That field belongs better into the `TracingActor`, I think.

src/som/interpreter/actors/Actor.java
```diff
@@ -190,12 +203,73 @@ protected void appendToMailbox(final EventualMessage msg) {
...
   }
 
-  protected static void handleBreakPoints(final EventualMessage msg, final WebDebugger dbg) {
-    if (VmSettings.TRUFFLE_DEBUGGER_ENABLED && msg.isMessageReceiverBreakpointSet()) {
+  protected static void handleBreakpointsAndStepping(final EventualMessage msg, final WebDebugger dbg, final Actor actor) {
+    if (!VmSettings.TRUFFLE_DEBUGGER_ENABLED) {
+      return;
+    }
+
+    synchronized (actor) {
```
smarr: why do you need to synchronize on the actor?

src/som/interpreter/actors/Actor.java
```diff
@@ -190,12 +203,73 @@ protected void appendToMailbox(final EventualMessage msg) {
...
+
+      if (msg.isStepToReceiver() || msg.isMessageReceiverBreakpointSet() || actor.isStepToNextTurn()) {
+        dbg.prepareSteppingUntilNextRootNode();
+        if (actor.isStepToNextTurn()) { // reset flag
+          actor.setStepToNextTurn(false);
+        }
+      }
+
+      if (msg.isStepToNextTurn()) {
+        prepareStepToNextMessage(msg, actor);
```
smarr: this is inconsistent with the handing of `isStepToReceiver()`. Do you really need a trivial `prepareStepToNextMessage()` helper method?
Same for the other things below.

src/som/interpreter/actors/Actor.java
```diff
@@ -190,12 +203,73 @@ protected void appendToMailbox(final EventualMessage msg) {
...
-    if (VmSettings.TRUFFLE_DEBUGGER_ENABLED && msg.isMessageReceiverBreakpointSet()) {
+  protected static void handleBreakpointsAndStepping(final EventualMessage msg, final WebDebugger dbg, final Actor actor) {
+    if (!VmSettings.TRUFFLE_DEBUGGER_ENABLED) {
+      return;
+    }
+
+    synchronized (actor) {
+
+      if (msg.isStepToReceiver() || msg.isMessageReceiverBreakpointSet() || actor.isStepToNextTurn()) {
+        dbg.prepareSteppingUntilNextRootNode();
```
smarr: why do you need two different things for `msg.isStepToReceiver()` and `msg.isMessageReceiverBreakpointSet()`? Isn't that really the same?

src/som/interpreter/actors/Actor.java
```diff
@@ -190,12 +203,73 @@ protected void appendToMailbox(final EventualMessage msg) {
...
+        if (actor.isStepToNextTurn()) { // reset flag
+          actor.setStepToNextTurn(false);
+        }
+      }
+
+      if (msg.isStepToNextTurn()) {
```
smarr: Why does the msg also have a `isStepToNextTurn()`? Isn't one enough? either on the actor or the message?

src/som/interpreter/actors/EventualMessage.java
```diff
@@ -33,22 +35,32 @@
    * (with or without resolver).
    */
   protected final boolean triggerPromiseResolverBreakpoint;
+  /**
+   * Indicates if a stepping operation has been activated for this message.
+   */
+  protected SteppingType steppingType;
```
smarr: Why do you need this here?
Why can't you just use the existing flags?

src/som/interpreter/actors/Actor.java
```diff
@@ -190,12 +203,73 @@ protected void appendToMailbox(final EventualMessage msg) {
...
+
+ }
+
+  /**
+   * Check callbacks for stop in the resolution of the promise.
+   */
+  private static void prepareCallbackForStepping(final EventualMessage msg, final WebDebugger dbg) {
+    boolean triggerStepping = false;
+    if (msg instanceof PromiseSendMessage) {
+      if (((PromiseSendMessage) msg).getPromise().isTriggerStopBeforeExecuteCallback()) {
```
smarr: Why is this different from how the call back is handled?

src/som/interpreter/actors/EventualMessage.java
```diff
@@ -233,13 +245,18 @@ public SPromise getPromise() {
   /** The callback message to be send after a promise is resolved. */
   public static final class PromiseCallbackMessage extends PromiseMessage {
     protected final SPromise promise;
+    /**
+     * The promise on which this callback is registered on.
+     */
+    protected final SPromise promiseWhereIsRegistered;
```
smarr: `promiseWhereIsRegistered` is not english, please fix

src/som/interpreter/actors/EventualMessage.java
```diff
@@ -341,4 +362,33 @@ public boolean isMessageReceiverBreakpointSet() {
   public void setIsMessageReceiverBreakpoint(final boolean triggerBreakpoint) {
     this.triggerMessageReceiverBreakpoint = triggerBreakpoint;
   }
+
+  public boolean isStepToReceiver() {
+    if (this.steppingType != null) {
```
smarr: why do you need the null check? it would return false, too?

smarr: And why do you need all these helpers anyway?
And I still don't see why you can't just reuse the flags we already had for the breakpoints.

src/som/interpreter/actors/EventualSendNode.java
```diff
@@ -183,24 +196,72 @@ protected void sendDirectMessage(final Object[] args, final Actor owner,
...
-          messageReceiverBreakpoint.executeCheckIsSetAndEnabled(), promiseResolverBreakpoint.executeCheckIsSetAndEnabled());
+          messageReceiverBreakpoint.executeCheckIsSetAndEnabled(), promiseResolverBreakpoint.executeCheckIsSetAndEnabled(), stepping);
+
       registerNode.register(rcvr, msg, rcvr.getOwner());
     }
 
+    /**
+     * Validates if the message has null resolver
+     * when a step-to-promise-resolution has been made, in that case
+     * none stepping strategy is returned.
```
smarr: I don't understand why that's needed. What is this used for?

src/som/primitives/actors/PromisePrims.java
```diff
@@ -144,17 +153,52 @@ protected final SPromise registerWhenResolved(final SPromise rcvr,
...
+
+      SteppingStrategy strategy = ActivityThread.steppingStrategy();
+      SteppingType stepping = null;
+      boolean stop = false;
+
+      if (strategy == null) {
+        return null;
+      }
+
+      if (strategy instanceof ToMessageReceiver) {
```
smarr: The same code is duplicated above in `validateResolverForStepping()`, why is this necessary?

src/som/primitives/actors/PromisePrims.java
```diff
@@ -200,21 +244,56 @@ protected final SPromise registerOnError(final SPromise rcvr,
...
```
smarr: And even more code duplication. Why is this needed?

src/som/primitives/actors/PromisePrims.java
```diff
@@ -278,10 +358,44 @@ protected final SPromise registerWhenResolvedOrError(final SPromise rcvr,
...
```
smarr: And more????

src/tools/debugger/FrontendConnector.java
```diff
@@ -267,6 +267,10 @@ public Suspension getSuspension(final long activityId) {
     return webDebugger.getSuspension(activityId);
   }
 
+  public Set<Class<? extends Tags>> getSourceTags(final SourceSection source) {
```
smarr: Please remove, it isn't used.

src/som/interpreter/actors/EventualSendNode.java
```diff
@@ -183,24 +196,72 @@ protected void sendDirectMessage(final Object[] args, final Actor owner,
...
+
+      SteppingStrategy strategy = ActivityThread.steppingStrategy();
+      SteppingType stepping = null;
+      boolean stop = false;
+
+      if (strategy == null) {
+        return null;
+      }
+
+      if (strategy instanceof ToMessageReceiver) {
```
smarr: it is really unclear to me why you first do a `instanceof` check and then use a polymorphic method that could do everything you need.

src/tools/debugger/WebDebugger.java
```diff
@@ -101,7 +101,7 @@ public void prepareSteppingAfterNextRootNode() {
     breakpoints.prepareSteppingAfterNextRootNode();
   }
 
-  Suspension getSuspension(final long activityId) {
+  public Suspension getSuspension(final long activityId) {
```
smarr: I doubt this change is still needed, please revert.

src/tools/debugger/WebDebugger.java
```diff
@@ -208,4 +208,12 @@ public static Gson createJsonProcessor() {
         registerTypeAdapterFactory(breakpointAF).
         create();
   }
+
+  public Set<Class<? extends Tags>> getTags(final SourceSection source) {
```
smarr: This method is not used, please remove.

src/tools/debugger/entities/SteppingType.java
```diff
@@ -104,11 +110,37 @@ public void process(final Suspension susp) {
     }
   },
 
+  @SerializedName("stepToMessageRcvr")
+  STEP_TO_RECEIVER_MESSAGE("stepToMessageRcvr", "Step to Msg Receiver", Group.ACTOR_STEPPING, "msg-open", new Class[] {EventualMessageSend.class}, new ActivityType[] {ActivityType.ACTOR}) {
```
smarr: I think we need to remove `new ActivityType[] {ActivityType.ACTOR}`, otherwise, you can't step from a `<-:` thats done by a process or thread to the receiver.

src/tools/debugger/entities/SteppingType.java
```diff
@@ -148,4 +180,4 @@ public void process(final Suspension susp) {
   }
 
   public abstract void process(Suspension susp);
-}
+}
```
smarr: It's unclear why you removed the linebreak.

src/tools/debugger/session/Breakpoints.java
```diff
@@ -180,9 +180,9 @@ public boolean evaluate() {
 
  public synchronized BreakpointEnabling<SectionBreakpoint> getReceiverBreakpoint(
       final FullSourceCoordinate section) {
-    return receiverBreakpoints.computeIfAbsent(section,
-        ss -> new BreakpointEnabling<SectionBreakpoint>(
-            new SectionBreakpoint(false, section, BreakpointType.MSG_RECEIVER)));
+   return receiverBreakpoints.computeIfAbsent(section,
```
smarr: why do you change the whitespace here?

tools/kompos/index.html
```diff
@@ -176,6 +176,10 @@
     cursor: ns-resize;
     background-color: #eeeeee;
   }
+
+  #msg-actions {
```
smarr: this is not used?

tools/kompos/index.html
```diff
@@ -240,6 +244,9 @@
     font-size: 75%;
     padding-top: 3ex;
   }
+  .activity-actions-msg {
```
smarr: This is not used?

tools/kompos/index.html
```diff
@@ -299,8 +311,8 @@
 
 <div id="code-views" class="card-columns"></div>
 
-<div class="card system-status" data-spy="affix">
-<div id="debugger" class="card-body">
+<div class="card system-status">
```
smarr: Why do you remove ` data-spy="affix"` here?

tools/kompos/index.html
```diff
@@ -282,6 +289,11 @@
   .trace-method {
     float:left;
   }
+    .div-scroll {
```
smarr: Why is this needed?

tools/kompos/index.html
```diff
@@ -400,12 +411,16 @@
...
-    <i id="icon-arrow-down"  class="fa fa-share" style="transform: rotate(90deg);"></i>
-    <i id="icon-arrow-right" class="fa fa-share"></i>
-    <i id="icon-arrow-left"  class="fa fa-reply" style="transform: scaleY(-1);"></i>
+    <i id="icon-play"         class="fa fa-play"></i>
+    <i id="icon-pause"        class="fa fa-pause"></i>
+    <i id="icon-stop"         class="fa fa-stop"></i>
+    <i id="icon-arrow-down"   class="fa fa-share" style="transform: rotate(90deg);"></i>
+    <i id="icon-arrow-right"  class="fa fa-share"></i>
+    <i id="icon-arrow-left"   class="fa fa-reply" style="transform: scaleY(-1);"></i>
+    <i id="icon-msg-open"     class="fa fa-envelope-open"></i>
```
smarr: I am not sure these icons are intuitive.

tools/kompos/src/messages.ts
```diff
@@ -81,6 +81,7 @@ export interface StoppedMessage {
   activityType:      ActivityType;
   text:              string;
   allThreadsStopped: boolean;
+  steppingCommands:  Boolean[];
```
smarr: Is this needed? It isn't used.

tools/kompos/src/ui-controller.ts
```diff
@@ -234,4 +234,5 @@ export class UiController extends Controller {
     this.vmConnection.sendDebuggerAction(step, act);
     this.view.onContinueExecution(act);
   }
+
```
smarr: Please avoid unnecessary whitespace changes.

tools/kompos/src/view.ts
```diff
@@ -861,4 +861,5 @@ export class View {
     const highlightedNode = $("#" + id + " .DbgCurrentNode");
     highlightedNode.removeClass("DbgCurrentNode");
   }
+
```
smarr: Please avoid unnecessary whitespace changes.

tools/kompos/index.html
ctrlpz: I removed it because it does not scroll the elements inside the div. I'm not sure why does not work, I read this property should lock the div in a specific area and enable scrolling. The solution I found was to add a css style to the div (.div-scroll), specifying a height and enabling scrolling vertically. I think this is needed because if we do not specify a fixed size for the debugger div, it can overlap the activity that is below it. As consequence, the user will not be able to press the buttons of the activity.

tools/kompos/src/messages.ts
ctrlpz: no, it isn't, I missed that in the rebase

tools/kompos/index.html
smarr: Could you be more specific? What doesn't scroll? I haven't noticed any problems with scrolling.

It was introduced to fix the debugger on the top right, more than a year ago: https://github.com/smarr/SOMns/commit/ca94f9deb9d83fde1656b17db8b10150cc42a95f

src/som/interpreter/actors/Actor.java
ctrlpz: Because in the `PromiseSendMessage` the method `getPromise` returns the promise to which that message was sent. In the case of `PromiseCallbackMessage` the method `getPromise` refers to the promise that is returned by the whenResolved. If we want to handle both cases with one approach, a solution will be to swap the semantics of `getPromise` and `getPromiseRegisteredOn` in `PromiseCallbackMessage`, then we could use `PromiseMessage.getPromise` method instead of the two separated approaches.

src/tools/debugger/entities/SteppingType.java
ctrlpz: I did a step-to-receiver-message in `<-:` and it works. Could you detail the scenario where does not work for me to reproduce?

src/tools/debugger/entities/SteppingType.java
ctrlpz: I think this happened in the rebase :(

src/tools/debugger/entities/SteppingType.java
smarr: With this approach, you can't step on messages that are send by a CSP process, a fork/join task, or a `Thread`.
And, I don't see the benefit of restricting it to actors.

src/som/interpreter/actors/Actor.java
ctrlpz: I did not realize I have it duplicated, the semantic I was thinking was not for this case, now I think it must be only in the actor

src/som/interpreter/actors/Actor.java
ctrlpz: why? but then the user will need to have enabled tracing always. Is tracing a functionality optional or mandatory to debug? I thought it was optional

smarr: Tracing is not optional, and it hasn't been optional for quite some time already (4-5month).
We need tracing data in the debugger for the visualization as well as the code panels.
The code that shows that is not optional is here:
https://github.com/smarr/SOMns/blob/master/src/som/vm/VmSettings.java#L58

src/som/interpreter/actors/Actor.java
ctrlpz: I saw you synchronized it in `getCurrentMessagesOrCompleteExecution`. I thought it was necessary also here, although I do not handle any message of the mailbox in this code. Could you tell me why do you use synchronized in the above method? I read  this commit but I think you don't explain this here https://github.com/MetaConc/SOMns/commit/86eb1f0e9c96c0bd57107155fdc0dbf4c05e7ac3

smarr: If you look at the code (the old code you referenced also needs to ensure that, a bit simpler, but not up-to-date) https://github.com/smarr/SOMns/blob/master/src/som/interpreter/actors/Actor.java#L292-L315, you will see that the state of the actor object is updated. The mailbox is exchanged, and the status of the actor is set to indicate that it is currently executing.

Since the mailbox is where all other threads/actors converge to sequentialize, it needs to be protected (as well as the state indicating whether an actor is executing or not).

It is unclear to my why you need to synchronize on the actor here, however.
Conceptually, there can only be a single thread executing an actor at a time.
So, all stepping operations that work on an actor are done by a single thread.
Consequently, I don't see a need why it needs to be synchronized.

src/som/interpreter/actors/Actor.java
```diff
@@ -271,7 +267,9 @@ protected void processCurrentMessages(final ActorProcessingThread currentThread,
     private void execute(final EventualMessage msg,
         final ActorProcessingThread currentThread, final WebDebugger dbg, final int i) {
       currentThread.currentMessage = msg;
-      handleBreakPoints(msg, dbg);
+      if (actor instanceof TracingActor) {
```
smarr: All other code uses checks like `if (VmSettings.ACTOR_TRACING)`, why do you do here something else?

src/som/interpreter/actors/Actor.java
```diff
@@ -404,4 +404,5 @@ public String getName() {
   public String toString() {
     return "Actor";
   }
+
```
smarr: Whitespace

src/som/primitives/actors/PromisePrims.java
```diff
@@ -88,11 +94,43 @@ public final SImmutableObject createPromisePair(final Object nil,
...
     @Override
     protected boolean isTaggedWithIgnoringEagerness(final Class<?> tag) {
-      if (tag == CreatePromisePair.class || tag == ExpressionBreakpoint.class) {
+      if (tag == CreatePromisePair.class || tag == ExpressionBreakpoint.class || tag == StatementTag.class) {
         return true;
       }
       return super.isTaggedWithIgnoringEagerness(tag);
     }
+
+    private boolean hasPromiseResolutionBreakpoint() {
```
smarr: Why is this code duplicated 4 times?

src/tools/debugger/entities/SteppingType.java
smarr: Could you please fix this?

src/som/primitives/actors/PromisePrims.java
ctrlpz: Because I think I need to enable the stepping for the primitives `createPromisePrim`, `whenResolved`, `whenResolvedOnError` and `onError`. And I don't see a common superclass for all of them where to put these two methods.

tools/kompos/index.html
ctrlpz: which icons do you suggest? I think will be better to differentiate them somehow from the ones of sequential stepping. Right now there are a lot of arrows for the buttons I think

PR #130, opened 2017-04-11
Fix perform:withArguments: primitive

daumayr:
Fixed perform: withArguments: primitive, previously the primitive was dead code.


smarr:
@daumayr thanks!
@cfscholl this fix is hopefully useful for you.

PR #131, opened 2017-04-16
Prepare settings and infrastructure for enabling automatic source formatting

smarr:
The goal is to have all Java code be formatted automatically.
This is to make code reviews simpler and avoid style discussions.

This is work in progress, need to find a good setting and solve the following todos:

- [x] review all changes for suitability of settings
- [x] fix overlong lines that lead to strange columnized formatting
- [x] add proper OSX/Linux support for auto-formatting with ant
- [x] enable format on safe for Eclipse project
- [x] add automatic formatting for TypeScript code: https://github.com/vvakame/typescript-formatter
- [x] make sure travis uses TS lint

@ctrlpz, I think, you wanted this for a long time.
@daumayr and @richard-roberts, think this will affect you...

This should only be merged when we can avoid issues with long-running branches.
Comments on desired style/settings welcome.


ctrlpz:
Yes, sounds good to have this!

smarr:
Formatting for TypeScript was introduced with #167.

PR #141, opened 2017-05-01
Fix race between symbol strings and trace data

smarr:
@daumayr this is more a TODO than and PR. Could you please have a look at this?

If I remember correctly, you introduced this originally.
And lately, I am having trouble with strings not yet being available when a trace is parsed.

Could you add some way of synchronizing the two?
I see multiple possible options.
One would be to have a id for the last symbol message a trace depends on, and use promises in the frontend to check that the symbols are already there.

Another option could be to expect a response for the symbol message and deal with the race in the backend.

At the moment, we have a similar problems with the activity definitions.
And I solve that in the frontend with promises: https://github.com/smarr/SOMns/blob/master/tools/kompos/src/ui-controller.ts#L105

### DIFF
--- a/src/tools/concurrency/ActorExecutionTrace.java
+++ b/src/tools/concurrency/ActorExecutionTrace.java
@@ -222,7 +222,7 @@ public class ActorExecutionTrace {
 
     // at the beginning of buffer, allows to track what was created/executed
     // on which thread, really cheap solution, timestamp?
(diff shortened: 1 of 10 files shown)


daumayr:
so the symbol message is processed after trace events needing that data?

smarr:
yes, exactly

PR #142, opened 2017-05-01
Channel, Process, and Fork/Join stepping

smarr:
Implement stepping on channel reads/writes and process creation, task creation, and return form activity.

This change also introduces stepping strategies to realize the stepping process.

- [x] added stepping into newly started process
- [x] add stepping to channel reader, from writer
- [x] add stepping to channel writer, from reader
- [x] add stepping to return from current activity (not applicable to actors)

- remove unused list of processes
- add `SomLanguage.getVM(rootNode)` for slow path use
- during debugging, fork/join threads need to track current task
- stepping types might be for specific types of activities only
- mark `#spawn:*` as statement to allow single stepping to it.

### DIFF
--- a/src/som/interpreter/SomLanguage.java
+++ b/src/som/interpreter/SomLanguage.java
@@ -3,6 +3,7 @@ package som.interpreter;
 import java.io.IOException;
 
 import com.oracle.truffle.api.CallTarget;
+import com.oracle.truffle.api.CompilerAsserts;
 import com.oracle.truffle.api.CompilerDirectives.CompilationFinal;
 import com.oracle.truffle.api.Truffle;
 import com.oracle.truffle.api.TruffleLanguage;
@@ -142,6 +143,11 @@ public final class SomLanguage extends TruffleLanguage<VM> {
 
   public VM getVM() {
     return vm;
+  }
+
+  public static VM getVM(final RootNode root) {
+    CompilerAsserts.neverPartOfCompilation("This is a simple hack to get the VM object, and should never be on the fast path");
+    return root.getLanguage(SomLanguage.class).getVM();
   }
 
   // Marker source used to start execution with command line arguments
--- a/src/som/interpreter/actors/Actor.java
+++ b/src/som/interpreter/actors/Actor.java
@@ -184,7 +184,7 @@ public class Actor implements Activity {
   }
 
   protected static void handleBreakPoints(final EventualMessage msg, final WebDebugger dbg) {
-    if (VmSettings.TRUFFLE_DEBUGGER_ENABLED && msg.isBreakpoint()) {
+    if (VmSettings.TRUFFLE_DEBUGGER_ENABLED && msg.isMessageReceiverBreakpointSet()) {
       dbg.prepareSteppingUntilNextRootNode();
     }
   }
--- a/src/som/interpreter/actors/Actor.java
+++ b/src/som/interpreter/actors/Actor.java
@@ -147,6 +147,13 @@ public class Actor implements Activity {
   @Override
   public long getId() { return 0; }
 
+  @Override
+  public void setStepToJoin(final boolean val) {
+    throw new UnsupportedOperationException(
+        "Return from activity, and step to join are not supported " +
+        "for event-loop actors. This code should never be reached.");
+  }
+
   /**
    * Send the give message to the actor.
    *
--- a/src/som/interpreter/actors/Actor.java
+++ b/src/som/interpreter/actors/Actor.java
@@ -12,7 +12,6 @@ import som.VM;
 import som.interpreter.objectstorage.ObjectTransitionSafepoint;
 import som.primitives.ObjectPrims.IsValue;
 import som.vm.Activity;
-import som.vm.ActivityThread;
 import som.vm.VmSettings;
 import som.vmobjects.SAbstractObject;
 import som.vmobjects.SArray.STransferArray;
@@ -330,8 +329,7 @@ public class Actor implements Activity {
     }
   }
 
-  public static final class ActorProcessingThread extends TracingActivityThread
-      implements ActivityThread {
+  public static final class ActorProcessingThread extends TracingActivityThread {
     public EventualMessage currentMessage;
 
     protected Actor currentlyExecutingActor;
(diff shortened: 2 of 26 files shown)


src/som/primitives/processes/ChannelPrimitives.java
```diff
@@ -249,6 +261,18 @@ protected boolean isTaggedWithIgnoringEagerness(final Class<?> tag) {
         return super.isTaggedWithIgnoringEagerness(tag);
       }
     }
+
+    private static boolean stopOnRead() {
```
smarr: I was planning to take care of these methods by actually integrating them into the breakpoint nodes. I think, it will be best to check the stepping strategy there.

The stepping strategy implementation itself is also something that probably can be solved simpler with a enum or so.

src/som/interpreter/SomLanguage.java
```diff
@@ -144,6 +145,11 @@ public VM getVM() {
     return vm;
   }
 
+  public static VM getVM(final RootNode root) {
```
smarr: "slow path" is supposed to mean: never part of JIT compiled code.

src/som/interpreter/actors/Actor.java
```diff
@@ -185,7 +191,7 @@ protected void appendToMailbox(final EventualMessage msg) {
   }
 
   protected static void handleBreakPoints(final EventualMessage msg, final WebDebugger dbg) {
-    if (VmSettings.TRUFFLE_DEBUGGER_ENABLED && msg.isBreakpoint()) {
+    if (VmSettings.TRUFFLE_DEBUGGER_ENABLED && msg.isMessageReceiverBreakpointSet()) {
```
smarr: *sigh* I think I have been asking for that change already a long time ago, no?
It would be really great if you could just split your work into smaller increments and we integrate stuff more often.
Avoids double work for both of us.

src/som/interpreter/actors/EventualMessage.java
```diff
@@ -323,15 +323,15 @@ public static long getCurrentExecutingMessageId() {
    * Indicates that execution should stop and yield to the debugger,
    * before the computed value is used to resolve the promise.
    */
-  public boolean isTriggerPromiseResolverBreakpoint() {
+  public boolean isPromiseResolverBreakpointSet() {
```
smarr: why 'trigger'? Because the flag triggers the breakpoint? hmmm. don't know.

src/tools/debugger/SteppingStrategy.java
```diff
@@ -0,0 +1,44 @@
+package tools.debugger;
+
+import som.vm.Activity;
+
+public abstract class SteppingStrategy {
```
smarr: Hm, we should look into that in more detail. I don't think you need a source section, tags, and nodes.

I don't need them for my breakpoints, and I don't see why yours would be so fundamentally different that you need all that. But perhaps I am missing something.

src/som/primitives/ActivitySpawn.java
```diff
@@ -36,31 +38,48 @@
...
     } else {
-      return new SomForkJoinTask(argArray);
+      return new SomForkJoinTask(argArray, stopOnRoot);
     }
   }
 
-  private static Process createProcess(final SObjectWithClass obj, final SourceSection origin) {
+  private static Process createProcess(final SObjectWithClass obj,
+      final SourceSection origin, final boolean stopOnRoot) {
     if (VmSettings.ACTOR_TRACING) {
```
smarr: yes, it should be renamed. But not in this PR.

src/tools/debugger/entities/BreakpointType.java
```diff
@@ -135,6 +135,15 @@ public void registerOrUpdate(final Breakpoints bps, final SectionBreakpoint bpIn
     }
   },
 
+  @SerializedName("activityOnExecBP")
+  ACTIVITY_ON_EXEC("activityOnExecBP", "On execution",
```
smarr: No, as the breakpoint definition here hopefully states, it is for all operations tagged with `ActivityCreation`. And currently, that's the `ActivitySpawn` primitives. So, processes, tasks, threads.
It doesn't really make sense for actors though.

src/tools/debugger/entities/SteppingType.java
```diff
@@ -30,44 +37,86 @@ public void process(final SuspendedEvent event) {
...
+  STEP_TO_RECEIVER_MESSAGE("todo",   "todo", Group.ACTOR_STEPPING, "arrow-right", null) { @Override public void process(final Suspension susp) { /* TODO */ } },
+  STEP_TO_PROMISE_RESOLUTION("todo", "todo", Group.ACTOR_STEPPING, "arrow-right", null) { @Override public void process(final Suspension susp) { /* TODO */ } },
+  STEP_TO_NEXT_MESSAGE("todo",       "todo", Group.ACTOR_STEPPING, "arrow-right", null) { @Override public void process(final Suspension susp) { /* TODO */ } },
+  STEP_RETURN_TO_PROMISE_RESOLUTION("todo", "todo", Group.ACTOR_STEPPING, "arrow-right", null) { @Override public void process(final Suspension susp) { /* TODO */ } };
 
   public enum Group {
     BASIC_CONTROLS("Basic Controls"),
     LOCAL_STEPPING("Local Stepping"),
-    ACTOR_STEPPING("Actor Stepping");
+    ACTIVITY_STEPPING("Activity Stepping"),
```
smarr: Activity Stepping is more general than Actor Stepping.
It local to one or two activities, step into activity and return from activity. Please see where `ACTIVITY_STEPPING` is actually used.

src/tools/debugger/session/Breakpoints.java
```diff
@@ -54,6 +54,7 @@
 
   /** Manually managed by us, instead of Truffle. */
   private final Map<FullSourceCoordinate, BreakpointEnabling<SectionBreakpoint>> channelOppositeBreakpoints;
+  private final Map<FullSourceCoordinate, BreakpointEnabling<SectionBreakpoint>> onExecutionBreakpoints;
```
smarr: It's for the breakpoints triggered when an activity starts executing, the `ACTIVITY_ON_EXEC` breakpoints/steps.
So, it's for processes, tasks, and threads.

PR #146, opened 2017-05-05
Abstract lexical constructs (preparation for Object Literals)

richard-roberts:
In order to enable a cleaner implementation of object literals I have generalised the behaviour for the nesting of lexical constructs. In particular, I have introduced a new `LexicalBuilder` superclass, of which `MethodBuilder` and `MixinBuilder` are subclasses.

LexicalBuilder builders can be nested arbitrarily. By relaxing the nesting in this way, it will be simpler to reuse `MixinBuilder`s to implement objects literals (issue #86, which requires `MixinBuilder`s to be nested inside of `MethodBuilder`s).

Beyond enabling a simpler implementation for object literals, these changes remove the need for builders to keep a reference to both the outer self (in the sense of the containing object) and the contextual self - now only the contextual self is needed. Similar changes have been implemented in the `LexicalScope` construct.

Finally, the introduction of the `SObjectWithContext` provides an single interface for interacting with VM objects that have a context. Currently the only implementor of this context is `SBlock`, but soon to be joined by `SObjectWithClass` to enable object literals.

### DIFF
--- a/src/som/compiler/LexicalBuilder.java
+++ b/src/som/compiler/LexicalBuilder.java
@@ -1,0 +1,82 @@
+package som.compiler;
+
+import som.interpreter.LexicalScope;
+
+public class LexicalBuilder {
+
+  private final LexicalBuilder outerBuilder;
+  private int                  levelsToNextMixin = -1;
+  private int                  levelsToNextMethod = -1;
+
+  public LexicalBuilder(final LexicalBuilder outerBuilder) {
+    this.outerBuilder = outerBuilder;
+  }
+
+  public LexicalScope getScope() {
+    return null;
+  }
+
+  public LexicalBuilder getContextualSelf() {
+    return outerBuilder;
+  }
+
+  private MixinBuilder bubbleToNextMixinOrNull() {
+    LexicalBuilder builder = getContextualSelf();
+    levelsToNextMixin = 0;
+
+    while (true) {
+
+      if (builder == null) {
+        return null;
+      }
+
+      if (builder instanceof MixinBuilder) {
+        return (MixinBuilder) builder;
+      }
+      builder = builder.getContextualSelf();
+
+      levelsToNextMixin += 1;
+    }
+  }
+
+  private MethodBuilder bubbleToNextMethodOrNull() {
+    LexicalBuilder builder = getContextualSelf();
+    levelsToNextMethod = 1;
+
+    while (true) {
+
+      if (builder == null) {
+        return null;
+      }
+
+      if (builder instanceof MethodBuilder) {
+        return (MethodBuilder) builder;
+      }
+
+      builder = builder.getContextualSelf();
+      levelsToNextMethod += 1;
+    }
+  }
+
+  public MixinBuilder getOuterSelf() {
+    return bubbleToNextMixinOrNull();
+  }
+
+  public MethodBuilder getNextMethod() {
+    return bubbleToNextMethodOrNull();
+  }
+
+  public int countLevelsToNextMixin() {
+    if (levelsToNextMixin == -1) {
+      bubbleToNextMixinOrNull();
+    }
+    return levelsToNextMixin;
+  }
+
+  public int countLevelsToNextMethod() {
+    if (levelsToNextMethod == -1) {
+      getNextMethod();
+    }
+    return levelsToNextMethod;
+  }
+}
--- a/src/som/compiler/MethodBuilder.java
+++ b/src/som/compiler/MethodBuilder.java
@@ -58,10 +58,8 @@ import som.vmobjects.SInvokable.SInitializer;
 import som.vmobjects.SSymbol;
 
 
-public final class MethodBuilder {
+public final class MethodBuilder extends LexicalBuilder {
 
-  private final MixinBuilder  directOuterMixin; // to get to an indirect outer, use outerBuilder
-  private final MethodBuilder outerBuilder;
   private final boolean       blockMethod;
 
   private final SomLanguage language;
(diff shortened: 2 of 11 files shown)


src/som/compiler/LexicalBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+
+  public LexicalScope getScope() {
+    return null;
+  }
+
+  public LexicalBuilder getContextualSelf() {
+    return outerBuilder;
+  }
+
+  private MixinBuilder bubbleToNextMixinOrNull() {
```
smarr: can we find a nicer word for `bubble`? Is 'outer' to overloaded? 
But it seems to be like outer could be the right word here?

src/som/compiler/LexicalBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+    LexicalBuilder builder = getContextualSelf();
+    levelsToNextMixin = 0;
```
smarr: variable not used?

src/som/compiler/LexicalBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+      if (builder instanceof MethodBuilder) {
+        return (MethodBuilder) builder;
+      }
+
+      builder = builder.getContextualSelf();
+      levelsToNextMethod += 1;
+    }
+  }
+
+  public MixinBuilder getOuterSelf() {
```
smarr: why not use this method directly to implement the functionality?

src/som/compiler/LexicalBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+    return bubbleToNextMixinOrNull();
+  }
+
+  public MethodBuilder getNextMethod() {
```
smarr: same thing here? why an extra method?

smarr: I don't like `next`. And, I don't like `method`.
Perhaps MethodBuilder?

src/som/compiler/LexicalBuilder.java
smarr: would it make sense to call it `getOuterMixinBuilder`?

src/som/compiler/LexicalBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+    return bubbleToNextMethodOrNull();
+  }
+
+  public int countLevelsToNextMixin() {
```
smarr: I'd prefer `get` over `count`.
Hide the implementation, also in the name...
Same for next method. 
And naming should be consistent with/parallel to above methods

src/som/compiler/MethodBuilder.java
```diff
@@ -177,17 +175,13 @@ public MethodScope getCurrentMethodScope() {
...
-  }
-
   // Name for the frameOnStack slot,
   // starting with ! to make it a name that's not possible in Smalltalk
   private static final String FRAME_ON_STACK_SLOT_NAME = "!frameOnStack";
 
   public Internal getFrameOnStackMarkerVar() {
-    if (outerBuilder != null) {
-      return outerBuilder.getFrameOnStackMarkerVar();
+    if (getNextMethod() != null) {
```
smarr: yeah, next really doesn't work here. (discussion above)

src/som/compiler/MethodBuilder.java
```diff
@@ -215,17 +209,17 @@ public boolean requiresContext() {
...
-      ctx = ctx.outerBuilder;
+      ctx = ctx.getNextMethod();
     }
     return ctx;
   }
 
   public boolean needsToCatchNonLocalReturn() {
     // only the most outer method needs to catch
-    return needsToCatchNonLocalReturn && outerBuilder == null;
+    return needsToCatchNonLocalReturn && getContextualSelf() instanceof MixinBuilder;
```
smarr: Do you mind if I change my mind on stuff?

We might indeed consider `isMixin()` and `isMethod()` methods to avoid all these `instanceof` tests.

src/som/compiler/MethodBuilder.java
```diff
@@ -360,22 +354,18 @@ public boolean isBlockMethod() {
...
   private int getContextLevel(final String varName) {
     if (locals.containsKey(varName) || arguments.containsKey(varName)) {
       return 0;
     }
 
-    if (outerBuilder != null) {
-      return 1 + outerBuilder.getContextLevel(varName);
+    MethodBuilder nextMethod = getNextMethod();
+    if (nextMethod != null) {
+      int nLevels = countLevelsToNextMethod();
```
smarr: I'd drop the `n` on the name. I am not a fan of 1-char abbreviations, except when I use them myself because of laziness ;)

src/som/compiler/MethodBuilder.java
```diff
@@ -521,30 +511,14 @@ public ReturnNonLocalNode getNonLocalReturn(final ExpressionNode expr,
...
-      }
-    } else {
-      return directOuterMixin;
-    }
-  }
-
   public ExpressionNode getOuterRead(final String outerName,
       final SourceSection source) throws MixinDefinitionError {
-    MixinBuilder enclosing = getEnclosingMixinBuilder();
+    MixinBuilder enclosing = getOuterSelf();
```
smarr: So, the whole notion of 'enclosing' was eliminated?
Not sure yet whether that's a good or bad thing.

Could you do a fuzzy search over the code to make sure we are consistent?
And, I want good definitions of `context` and `outer`, and all relevant methods should have a comment, to reflect the definitions. Pretty please :)
I know, I am asking more from you than from myself. But, yeah, since we now have the opportunity to do it right, we might be nicer to our future selfs then I was...

src/som/compiler/Parser.java
```diff
@@ -1486,7 +1485,10 @@ private ExpressionNode nestedBlock(final MethodBuilder builder) throws ProgramDe
       blockPattern(builder);
     }
 
-    String outerMethodName = stripColons(builder.getOuterBuilder().
+    if (builder.getNextMethod() == null) {
+      throw new java.lang.NullPointerException("The outer context of a block should never be null");
```
smarr: Why fully qualified?
And also why `NullPointer`? Isn't `IllegalState` or something more appropriate?

src/som/interpreter/LexicalScope.java
```diff
@@ -18,24 +18,54 @@
...
+
+  public LexicalScope(final LexicalScope outer) {
+    this.outer = outer;
+  }
+
+  public LexicalScope getContextualScope() {
+    return outer;
+  }
+
+  public MixinScope getOuterScope() {
```
smarr: Same naming discussion as with the builders.
This needs to parallel them exactly.

src/som/interpreter/LexicalScope.java
```diff
@@ -18,24 +18,54 @@
 
 public abstract class LexicalScope {
 
+  private final LexicalScope outer;
+
+  public LexicalScope(final LexicalScope outer) {
```
smarr: is `outer` here the correct term?

src/som/interpreter/LexicalScope.java
```diff
@@ -18,24 +18,54 @@
...
+    this.outer = outer;
+  }
+
+  public LexicalScope getContextualScope() {
+    return outer;
```
smarr: might need to be `context`?
At least that's what the method name suggests?
Also `context` vs. `contextual`. Please stick with one.

src/som/interpreter/LexicalScope.java
```diff
@@ -18,24 +18,54 @@
...
-    public MixinScope(final MixinScope outerMixin) {
-      this.outerMixin = outerMixin;
-    }
-
-    public MixinDefinition getOuter() {
-      if (outerMixin != null) {
-        return outerMixin.mixinDefinition;
-      }
-      return null;
+    public MixinScope(final LexicalScope contextualScope) {
```
smarr: context vs. contextual?

src/som/vmobjects/SBlock.java
```diff
@@ -78,3 +80,4 @@ public ForeignAccess getForeignAccess() {
     return SBlockInteropMessageResolutionForeign.ACCESS;
   }
 }
+
```
smarr: why the extra whitespace?

src/som/vmobjects/SObjectWithContext.java
```diff
@@ -0,0 +1,11 @@
+package som.vmobjects;
+
+import com.oracle.truffle.api.frame.MaterializedFrame;
+
+public interface SObjectWithContext {
+
+  MaterializedFrame getContext();
+  Object getContextualSelf();
```
smarr: shouldn't that be outer?
Might be a good place for the definitions?

smarr:
Also, good that you split that out as a separate PR.
It's already a lot of changes.
So, decoupling them, makes it much easier to review.

PR #147, opened 2017-05-05
Fix Threads for Tracing, and add STM stepping support

smarr:
- add transaction stepping
  - next transaction
  - step to commit
  - step after commit

Add the notion of dynamic concurrent entity scope for stepping buttons.
This allows us to enable buttons for transactions or under held locks only.
Similar to the 'current activity' notion, but as a stack of dynamic activations.

Turn threads also into tasks on a fork/join pool to have a uniform representation and simplify tracing and debugging infrastructure.

### DIFF
--- a/src/tools/debugger/entities/SteppingType.java
+++ b/src/tools/debugger/entities/SteppingType.java
@@ -7,9 +7,11 @@ import tools.concurrency.Tags;
 import tools.concurrency.Tags.ActivityCreation;
 import tools.concurrency.Tags.ChannelRead;
 import tools.concurrency.Tags.ChannelWrite;
+import tools.debugger.SteppingStrategy.AfterCommit;
 import tools.debugger.SteppingStrategy.IntoSpawn;
 import tools.debugger.SteppingStrategy.ReturnFromActivity;
 import tools.debugger.SteppingStrategy.ToChannelOpposite;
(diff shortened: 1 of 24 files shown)


src/tools/debugger/entities/SteppingType.java
```diff
@@ -156,19 +189,29 @@ private static void completeHandleFrameSkip(final Suspension susp) {
       If no tags are given, it is assumed the operation is always valid. */
   public final Class<? extends Tags>[] applicableTo;
 
+  /** Stepping operation is only available when in the dynamic scope of the given entity.
+      If no entity types are given, it is assumed the operation is always valid. */
+  public final EntityType[] inScope;
```
smarr: It is for operations that are scoped by transactions, or object mutexes, or other dynamic scopes. At the moment, I only have these two examples.

smarr: Technically, we could see a turn also as a dynamic scope, but, it's not necessary since all operations executed by an actor are executed within a turn. So, it's somewhat redundant.
If there would be any code that's executed outside the scope of turns, it would be necessary.
One example would be actor models that handle message receiving differently and do not have a notion of a turn perhaps.

ctrlpz: ok, I understand the idea, thanks

PR #153, opened 2017-05-11
Refactor Handling of Stepping

smarr:
The goal of this change is to simplify the implementation of stepping and unify it with breakpoints where possible.
Thus, in most cases, a potential breakpoint is also a stepping target.

This means, a `BreakpointNode` should be able to do the test whether a relevant stepping strategy is activated.

This change also removes the boilerplate code from `SteppingStrategy`. There doesn't seem to be a need for this complexity. A simple check for the `SteppingType` seems sufficient.

This PR also adds the missing *step-to-promise-resolver* stepping.


src/tools/debugger/entities/SteppingType.java
```diff
@@ -148,34 +137,51 @@ public void process(final Suspension susp) {
...
-      susp.getActivityThread().setSteppingStrategy(new ToMessageReceiver());
-      }
-    },
+      susp.getActivityThread().setSteppingStrategy(this);
+    }
+  },
+
+  @SerializedName("stepToPromiseResolver")
+  STEP_TO_PROMISE_RESOLVER("stepToPromiseResolver", "Step to Promise Resolver",
+      Group.ACTOR_STEPPING, "msg-white",
```
ctrlpz: I should probably find more intuitive icons for each stepping type, they have to be different for each command.

src/tools/debugger/entities/SteppingType.java
```diff
@@ -148,34 +137,51 @@ public void process(final Suspension susp) {
...
       susp.getEvent().prepareContinue();
-      susp.getActivityThread().setSteppingStrategy(new AfterCommit());
+      susp.getActivityThread().setSteppingStrategy(this);
     }
   },
 
   @SerializedName("stepToMessageRcvr")
-  STEP_TO_RECEIVER_MESSAGE("stepToMessageRcvr", "Step to Msg Receiver", Group.ACTOR_STEPPING, "msg-open", new Class[] {EventualMessageSend.class}) {
-    @Override public void process(final Suspension susp) {
+  STEP_TO_MESSAGE_RECEIVER("stepToMessageRcvr", "Step to Msg Receiver",
```
ctrlpz: thanks for renaming this command, I realize I had a typo there

src/tools/debugger/frontend/Suspension.java
```diff
@@ -145,12 +148,21 @@ public void suspend() {
...
-          strategy.handleResumeExecution(activity);
-          strategy.handleMessageToNextTurn(activity);
-        }
 
+        if (!continueWaiting) {
+          if (activityThread.isStepping(SteppingType.RETURN_FROM_ACTIVITY)) {
+            activity.setStepToJoin(true);
+          } else if (activityThread.isStepping(SteppingType.STEP_TO_NEXT_TURN)) {
+            activity.setStepToNextTurn(true);
+          } else if (activityThread.isStepping(SteppingType.RETURN_FROM_TURN_TO_PROMISE_RESOLUTION)) {
```
ctrlpz: I think I understand what you mean here, is similar to the step-to-next-turn command in the sense that should be able to trigger from anywhere in the code. And I only have it implemented for the `SendNode` and the primitives of `PromisePrims`

src/tools/debugger/session/Breakpoints.java
```diff
@@ -100,11 +109,8 @@ public synchronized void addOrUpdateBeforeExpression(final SectionBreakpoint bId
 
   public synchronized void addOrUpdateAfterExpression(final SectionBreakpoint bId) {
     // TODO: does this work???
```
ctrlpz: what happens here? I saw that is used for transactions and fork/join but don't we have tests for them?

src/tools/debugger/nodes/BreakpointNode.java
```diff
@@ -12,16 +13,16 @@
...
-  protected BreakpointNode(final BreakpointEnabling<?> breakpoint) {
+  protected BreakpointNode(final BreakpointEnabling breakpoint) {
     this.breakpoint = breakpoint;
   }
 
   @Specialization(assumptions = "bpUnchanged", guards = "breakpoint.isDisabled()")
   public final boolean breakpointDisabled(
       @Cached("breakpoint.getAssumption()") final Assumption bpUnchanged) {
-    return false;
+    return ActivityThread.isSteppingType(breakpoint.getSteppingType());
```
ctrlpz: then the disabled state of the breakpoint will depend on the stepping. I think is a good approach. It fits better in this location than what I implemented, I think.

src/tools/debugger/session/Breakpoints.java
smarr: That's an old comment, I think it mostly works.
But no, it isn't tested.

smarr:
The whole refactoring hinges on the realization that there is a mapping between breakpoints and stepping operations.

It is important to realize however that there is connection between breakpoints and stepping targets.
I think, all breakpoints should be possible targets for stepping operations.
And if only for something as generic as 'next concurrency operation'.
We don't have that yet. Especially most of the 'before' breakpoints are excluded.

Sometimes stepping operations also don't use the breakpoints for what should really be the same source location.
So, there aught to be more cleanup done here.
And for instance, for the thread-based stuff, I haven't implemented any stepping yet, because it's just more of the same.

PR #164, opened 2017-06-25
Simplify Object Model and add StorageAccessor

smarr:
- move variability into dispatch chains and remove read/write nodes
  - the goal here is to have one less indirection/moving part
    didn't seem necessary anymore, and just made the model more complex
  - prim (set/setOrUnset), object, unwritten variability is now in
    dispatch chains directly
- do accessing of the slot's memory in a new storage accessors
  - there is a fixed number of storage accessors preallocated based on
    the fields in S*Object
  - the goal is to have more control over the unsafe accesses and try
    alternative accessors for the SubstrateVM
- expose CheckSObject guard and use it for cast
- correct naming and file structure
- move storage handling and primitiveUsedMap handling from SObject to
  storage accessors


smarr:
@richard-roberts Would be great if you could have a look at this, and potential review the change.
It's probably going to be interesting for you in the long term.

@daumayr, I am not sure, no way to confirm, I think, but I might have fixed our long-standing race.
There was a race for updating the object layouts. We tried to initialize a previously unwritten field, or generalize it multiple times.

smarr:
The changes to the object model are merged now, but post-merge reviews are still welcome.

PR #169, opened 2017-06-28
Model promise BPs solely with onResolver and onResolution

smarr:
- this remove the whole notion of explicit promises
- remove notion of triggerResolutionBreakpointOnUnresolvedChainedPromise, not needed already models with onResolution
- remove notion of triggerExplicitPromiseResolverBreakpoint, not needed already modeled with onResolver

- make AbstractPromiseResolutionNode quaternary to take both breakpoint flags

- add haltOnPromiseMessageResolution() to eventual message to be able to test at the point where the callback messages are received and we need to trigger resolution breakpoints. It is separate from haltOnResolution() to not trigger it too early just for any message that has a promise with the flag set, we only need it for the PromiseMessage resulting from the first message

- eventual messages have now also haltOnReceive(), haltOnResolver(), and haltOnResolution()


src/som/interpreter/actors/ResolvePromiseNode.java
```diff
@@ -23,19 +23,17 @@
...
+    if (haltOnResolver || promise.haltOnResolver()) {
       haltNode.executeEvaluated(frame, result);
     }
-    if (promise.isExplicitPromise() && promise.isTriggerPromiseResolutionBreakpoint()) {
-      breakpointOnResolution = promise.isTriggerPromiseResolutionBreakpoint();
-    }
 
-    resolvePromise(Resolution.SUCCESSFUL, resolver, result, breakpointOnResolution);
+    resolvePromise(Resolution.SUCCESSFUL, resolver, result,
+        haltOnResolution || promise.haltOnResolution());
```
ctrlpz: As I understand when a promiseResolutionBreakpoint is set on a `createPromisePair` primitive for triggering the breakpoint, the implementation idea is the same: check a flag on the promise. With this new changes you reused the flag for triggering a promise resolution breakpoint, that you call `haltOnResolution`. I think it makes sense, I did not see that variant in that moment.

src/som/interpreter/actors/EventualMessage.java
```diff
@@ -323,24 +331,40 @@ public static EventualMessage getCurrentExecutingMessage() {
 
   /**
    * Indicates that execution should stop and yield to the debugger,
-   * before the computed value is used to resolve the promise.
+   * before the message is processed.
+   */
+  public boolean haltOnReceive() {
+    return haltOnReceive;
```
smarr: Could you be more specific why this is confusing?
`isHaltOnReceive` doesn't read well, "is halt" isn't really English.
Personally, I don't see this as an issue, but then again, I have used languages where it is more common to name accessors like the fields they access.

ctrlpz: for me is confusing when reading the code, I have not seen this in Java before, aren't we following Java conventions? maybe `isEnabledHaltOnReceive` is better English?

smarr: You haven't answered my question. So, why is this confusing?
And which conventions. Do you have a specific link I should be looking at?
I am not a Java native.

ctrlpz: you can read here: https://docs.oracle.com/javase/tutorial/javabeans/writing/properties.html
Also in the JDK classes you can see that Java programmers follow this naming convention. For example: [Container.java](https://docs.oracle.com/javase/7/docs/api/java/awt/Container.html), [ArrayList.java](https://docs.oracle.com/javase/7/docs/api/java/util/ArrayList.html)

smarr: You still haven't answered my question. Is this on purpose?

smarr: The link isn't really a style guide.
Either way, `isEnabledHaltOnReceive()` is much longer, and not much more grammatical to my ears.

If you answer my question why it is confusing, I am considering to name it `getHaltOnReceive()`.

ctrlpz: Is confusing because I have not seen it before. In Java is not common to use that naming you have used, that's why I sent you some links to JDK classes for you to see the pattern of `isPropertyName`. In your version, the only difference between variable and method name is the parenthesis.

smarr: So, is `getHaltOnReceive()` acceptable?

ctrlpz: yes, I think is acceptable

Issue #108, opened 2017-02-16
Assignment isn't parsed correctly, but as keyword


salenaer:
when writing the target variable and the assignment operator without space this is read as method: =
for example 
arg:=value
(actor<-: message) whenResolved: [:arg | value:= arg].

error
value: tried to parse primary expression found equal.

smarr:
Might be fixable, but requires lookahead on parsing keywords.
Also, switching to proper Newspeak setter sends (#7) would make this unnecessary.

smarr:
Issue solved/avoided with #170.

PR #170, opened 2017-07-11
Adopt Newspeak Setter Sends instead of Assignment Syntax

smarr:
This PR solves #7.

To be Newspeak spec compliant, we need to move from the classic `var := 42` assignment syntax to the Newspeak setter send syntax where we got `var:: 42` instead.
The main motivation for the change is compliance and the ability to load Newspeak code at some point directly.

The changes in the parser and lexer remove the previous use of an `Assign` symbol, and instead, following the Newspeak spec, lex a `SetterKeyword`. This makes sure that there can't be whitespace between the two colons.

##### SOMns Code Transformation

To adapt existing code, I used the following:

```
find . -name '*.som' -exec sed -i.bak 's/\ :=\ /:: /g' {} \;
```

Note that previously, multiple white spaces where allowed before the assignment `:=`.
To fix up parsing issues, and preserve alignment, I used this (actually I did search replace of the equivalent in my editor, so this is untested):

```
find . -name '*.som' -exec sed -i.bak 's/\ ::\ /::  /g' {} \;
```


smarr:
This solves also the issue in reported in #108. It solves it in the sense that this syntactic ambiguity is gone.

PR #173, opened 2017-07-18
Array literals implementation

richard-roberts:
A revamp of #100 to provide support array literals (see #83). With these changes, Newspeak's tuple literals are parsed as an ~Generic~ArrayLiteralNode. The node is instantiated with the list of expressions declared in the tuple, each of which are evaluated and stored into a new `Object[]`. If each of the expressions are evaluated to the same type, then the `Object[]` is replaced with a more specialised array (`long[]`, `double[]`, etc). The result of executing the node is an instance of `SMutableArray` that encapsulates the array of evaluated expressions.

Parser changes and a basic test has also been added.


src/som/interpreter/nodes/literals/GenericArrayLiteralNode.java
```diff
@@ -0,0 +1,186 @@
...
+  @ExplodeLoop
+  protected SMutableArray evaluatedExpressions(final VirtualFrame frame) {
+    Object[] arr = new Object[expressions.length];
+    for (int i = 0; i < expressions.length; i++) {
+      arr[i] = expressions[i].executeGeneric(frame);
+    }
+
+    GenericArrayLiteralNode replacedNode = this;
+
+    if (allComponentsAreBoolean(arr)) {
```
smarr: Hm, I am not sure this is the best way of doing it. Did you see [`evaluateFirstDetermineStorageAndEvaluateRest(.)`](https://github.com/smarr/SOMns/blob/master/src/som/primitives/arrays/ArraySetAllStrategy.java#L109)? I would try to stick to existing approaches, and possibly even reuse the code.
Which might also mean that we don't need a separate class for each type of array literal.

src/som/interpreter/nodes/literals/GenericArrayLiteralNode.java
```diff
@@ -0,0 +1,186 @@
...
+
+    @Override
+    @ExplodeLoop
+    protected SMutableArray evaluatedExpressions(final VirtualFrame frame) {
+      boolean[] arr = new boolean[expressions.length];
+      for (int i = 0; i < expressions.length; i++) {
+        try {
+          arr[i] = expressions[i].executeBoolean(frame);
+        } catch (UnexpectedResultException e) {
+          throw new RuntimeException("Expression " + i + " of this BooleanArrayLiteralNode did not evaluate to a double");
```
smarr: Ehm, ok?
That's not a semantically correct way to handle it :)
You can cheat, as long as you don't get caught :)

richard-roberts:
I've changed how the specialisation is done and now use a (simpler) specialise method instead of the separate classes.

I considered reusing behaviour from [evaluateFirstDetermineStorageAndEvaluateRest](https://github.com/smarr/SOMns/blob/master/src/som/primitives/arrays/ArraySetAllStrategy.java#L109) as @smarr suggested, but I would need to wrap the array literal's expressions into an `SBlock` to do so. It seemed like more computation than was necessary, so instead I've designed the specialisation based off of that array strategy.

smarr:
@richard-roberts, I took the liberty to rebase the branch on `dev`, which is the development branch, and the next major release.

Further, I changed the symbols to use the names from the Newspeak spec.
This is currently a little odd, and inconsistent with the rest. But, I think, we better start with something to avoid being completely inconsistent. So, `{` and `{}` are now `LCurly` and `RCurly`.

I am also thinking of squashing some of the inbetween commits, if that's ok with you.

src/som/interpreter/nodes/literals/ArrayLiteralNode.java
```diff
@@ -0,0 +1,81 @@
...
+    for (int i = 0; i < expressions.length; i++) {
+      evaluated[i] = expressions[i].executeGeneric(frame);
+    }
+    return evaluated;
+  }
+
+  @ExplodeLoop
+  private boolean headSameTypeAsTail(final Object[] evaluated) {
+    for (int i = 1; i < evaluated.length; i++) {
+      if (!(evaluated[i].getClass().equals(evaluated[0].getClass()))) {
```
smarr: Comparing the classes here does unfortunately not really work, since we have a wide range of classes floating around, not just primitives and objects.

smarr:
@richard-roberts please have a look at my changes.
I think, I didn't explain very well what my problem was with your specialization logic.

Specializations need to form a minimal state machine, which reaches a stable point as fast as possible. Each state in the state machine is one specialization. The state transitions are defined by the 'guards'.

With my implementation, we now got a state machine which looks like this:

```
(Uninit)
  -> all long -> (Longs)      -\
  -> all boolean -> (Booleans) -\
  -> all double -> (Doubles)   -|-> some object -> (Objects)
  -> all nil -> (Empty)        -/
  -----------------------------/
```

Further, the implementation does ideally only allocate the array that is needed, so, I speculate on everything being the same type as the first element. You checked that somehow, but there was always an Object[] if I am not mistaken.

And finally, I moved the implementation into `ArraySetAllStrategy` since we got already more or less the same there anyway.
It's easier to keep these implementations aligned if they reside in the same file.
While they are not identical, they are pretty close.
They should be identical, but I am a little to lazy for that...
Yeah, and if Java would be a better language, we could avoid much code duplication, too.

Hope this makes somewhat sense.

smarr:
I adapted code to use array literals where it makes sense.
The benchmarks look fine, too: http://somns-speed.stefan-marr.de/changes/?tre=10&rev=778c1f663758733d0accb3933abb1b2e2744440c&exe=19&env=1

smarr:
One more todo item for this: the dynamic metric tool needs to be taught about array literals to count their allocation

core-lib/TestSuite/TestRunner.som
```diff
@@ -89,6 +89,7 @@ class TestRunner usingPlatform: platform = (
                'DoubleTests',
                'IntegerTests',
                'StringTests',
+               'ArrayLiteralTests',
```
smarr: Done

smarr:
Ok, from my perspective, this would be done and ready to be merged.
@richard-roberts, I pushed the changes that I'd like to merge here: https://github.com/smarr/SOMns/tree/ns-array-literals-final

The final diff is identical to the current changes on your branch, but I simplified the history to focus on the essence of the changes. Will make our life easier to figure out issues in the future.

If that all works for you, I'll merge things as soon as possible.
Thanks.

src/som/interpreter/nodes/literals/ArrayLiteralNode.java
```diff
@@ -1,81 +1,158 @@
...
-    return new SMutableArray(executeAndSpecialize(frame), Classes.arrayClass);
+  private static final class Objects extends Specialized {
+    private Objects(final ExpressionNode[] expressions, final SourceSection source) {
+      super(expressions, source);
+    }
+
+    @Override
+    protected Object executeSpecialized(final VirtualFrame frame) {
+      return ArraySetAllStrategy.evalForRemaining(frame, expressions,
+          new Object[expressions.length], SArray.FIRST_IDX);
```
richard-roberts: Is there ever a case where `Array.FIRST_IDX` will not be zero?

richard-roberts:
@smarr thanks for picking up the slack and pushing this PR along. I've looked over the changes and checked the benchmarks -  let's merge this thing!

src/som/interpreter/nodes/literals/ArrayLiteralNode.java
smarr: That's an artifact of my work on metaobject protocols, for those, I put some hidden things into arrays, then it wasn't 0. At the moment it is, but I kept it around for the possibility that I might need it one day. Not sure it's a great idea.

PR #179, opened 2017-08-01
Fix turn stepping operations, and introduce better actor testing framework

smarr:
#### Turn Stepping Operations

Actor turn stepping operations needs to resume execution of current actor.
This includes step to next turn, and return from turn to resolution.
Without resuming execution of the current actor, the behavior in the debugger is really unintuitive. It says 'step to next turn', but it doesn't actually do that. It just did a step-over, which I don't understand why that would be desirable.

#### Add new way of writting stepping tests

The main benefit of this approach is that we can handle non-deterministic stepping results easily and don’t need to do it in the assertions manually.

This new stepping.ts introduces the data format for the tests, and it is currently used in actor.ts to generate tests.

The basic idea is that an initial set of breakpoints results in a single stop, from which we then can take separate steps, which each lead to n predetermined stops at given positions.

Note that the system uses aliases for the actors instead of ids which are fragile. It determines the activity id by matching it against the expected names used in the steps. So, it discovers the mapping at run time automatically. This works, as long as we introduce activities only gradually.

#### Other changes

- make Kompos tests run on more than one thread
- simplify actor demo for presentation
- don't allow stepping into creation of actors, doesn't make sense, they are second-class entities only

@ctrlpz please add this PR on your todo list for after you're done with your current work.
This is stuff that's relevant for you, and I want you to have a look at it.
But, I'll already merge the changes to be able to make progress on other things.

### DIFF
--- a/core-lib/demos/ActorDemo.som
+++ b/core-lib/demos/ActorDemo.som
@@ -9,23 +9,7 @@ class ActorDemo usingPlatform: platform = Value (
     )
   )
 
-  public woResolver = (
-    | p |
-    p:: (actors createActorFromValue: Result) <-: new.
-    p <-: get.
-    'after get send, sequential' println.
-    ^ 0
-  )
-
-  public wResolverNoCallback = (
-    | p p2 |
-    p:: (actors createActorFromValue: Result) <-: new.
-    p2:: p <-: get.
-    'after get send, sequential' println.
-    ^ 0
-  )
-
-  public wResolverAndCallback = (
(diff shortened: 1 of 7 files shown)


smarr:
The tests still contain race conditions because `pingpong.som` has parallel behavior that is not accounted for. Generally, `pingpong.som` should not be used for theses tests, it is too complex and I don't know what's going on there. We should use small simple programs to test the debugger, where we can write deterministic tests.

PR #186, opened 2017-08-31
Update Truffle to >0.27 and mx

smarr:
This PR updates Truffle.

For the new interop tests it only does the bare minimum and disables some tests where the engine is disposed, which might be fixed when using Truffle SDK.

Note, this PR does not yet contain any changes to handle newly introduced deprecations.


smarr:
The used Graal version is known to cause an assertion failure in `Savina.UnbalancedCobwebbedTree`.

PR #188, opened 2017-10-06
Fix dynamic metrics tool on core-lib/Benchmarks/All.ns

fredreichbier:
Hi,

these two commits fix running ``./som -G -dm ./core-lib/Benchmarks/All.ns`` for me: Previously, it failed with a ``java.lang.ArrayIndexOutOfBoundsException`` evoked by ``MetricsCsvWriter`` and because of an unhandled error in ``VectorTest``.

Please note that commented out the ``PageRank`` and ``GraphSearch`` benchmarks in my tests because they take a really long time on my machine (and the latter runs out of memory) :-)

Best Regards

Friedrich


smarr:
Thanks a lot for looking into this and fixing it.
Changes look good to me.
Just one small thing, since you changed `LanguageFeatures.ns`, we need updated test data for the dynamic metrics tests.

Could you please run `tests/dym/test.sh update` and commit the updated expected test results?

smarr (approved):

fredreichbier:
Cool! I just committed the updated expected test results, but some tests still seem to be failing :-/ Are these related to my changes?

smarr:
Those are unfortunately flaky tests. There are race conditions which haven't been fixed :(

Unrelated to your changes.

And thanks again for fixing small things.
Those changes look minor but unfortunately rarely get done.
So, thanks a lot!

PR #189, opened 2017-10-10
Outer sends need to check both super classes and mixins

ghost:
Classes created via mixin application should preserve access to variables and methods bound in enclosing scopes. 

Before these changes classes created via mixin application were not able to preserve this access; in particular, SOMns would fail to specialize `ResolvingImplicitReceiverSend` when one of the applied mixin's methods accessed something defined in its enclosing activation. 

For example in the following program (before these changes) SOMns would fail to specialize the implicit request for `bar` (from inside `A.foo`):

```
class MixinAppTest = ()(
  
  public class A = ()(
    public foo = ( ^ bar )
  )

  public class B = Object <: A ()()
  public bar = ( ^ 1 )

  public main: args = (
    (B new) foo.
  )
)
```

These changes enable SOMns to correctly resolve implicit requests when made by the methods of an applied mixin. In particular, logic has been added to `OuterObjectRead.getEnclosingClass(MixinDefinitionId)` to make this possible. As before, each super class (from shallowest to deepest) is checked to see if it matches the parameter `MixinDefinitionId`. If no viable super class is found, then each of the mixin's applied to each super class (again from shallowest to deepest) are checked to see if they match the parameter `MixinDefinitionId`. In both cases, the first superclass to meet the condition is selected as the class who should resolve the outer read. 

With the above changes, SOMns is able to recognise that the `B` contains within it a mixin that is based on `A` and, consequently, can specialize the implicit request for `bar` with `B` as the receiver. Therefore, with these changes SOMns can preserve access to variables and methods bound in enclosing scopes for the methods of applied mixins.

### DIFF
--- a/src/som/interpreter/nodes/OuterObjectRead.java
+++ b/src/som/interpreter/nodes/OuterObjectRead.java
@@ -81,8 +81,25 @@ public abstract class OuterObjectRead
 
   public abstract Object executeEvaluated(Object receiver);
 
+  /**
+   * The enclosing class can be either the class of the given receiver
+   * or any of its superclasses. The receiver can denote either a class
+   * or a mixin and, consequently, we need to check both the inheritance
+   * chain and any mixins applied to that class (or one of its superclasses).
+   * The class and its superclasses are queried first, and if no
+   * class was found then all mixins applied along the inheritance chain are
+   * queried.
+   *
+   * Note that this method should always find a class for the given receiver
+   * and that the return value must be either the receiver's class or one its
+   * superclasses (and not the class of a mixin).
+   */
   protected final SClass getEnclosingClass(final SObjectWithClass rcvr) {
-    SClass lexicalClass = rcvr.getSOMClass().getClassCorrespondingTo(mixinId);
+    SClass rcvrClass = rcvr.getSOMClass();
+    SClass lexicalClass = rcvrClass.lookupClass(mixinId);
+    if (lexicalClass == null) {
+      lexicalClass = rcvrClass.lookupClassWithMixinApplied(mixinId);
+    }
     assert lexicalClass != null;
     return lexicalClass;
   }
@@ -90,7 +107,7 @@ public abstract class OuterObjectRead
   protected static final SClass getEnclosingClassWithPotentialFailure(
       final SObjectWithClass rcvr,
       final int superclassIdx) {
-    SClass lexicalClass = rcvr.getSOMClass().getClassCorrespondingTo(superclassIdx);
+    SClass lexicalClass = rcvr.getSOMClass().lookupClass(superclassIdx);
     return lexicalClass;
   }
 
--- a/src/som/interpreter/objectstorage/ClassFactory.java
+++ b/src/som/interpreter/objectstorage/ClassFactory.java
@@ -6,6 +6,7 @@ import java.util.HashSet;
 import com.oracle.truffle.api.CompilerAsserts;
 
 import som.VM;
+import som.compiler.MixinBuilder.MixinDefinitionId;
 import som.compiler.MixinDefinition;
 import som.compiler.MixinDefinition.SlotDefinition;
 import som.interpreter.nodes.dispatch.Dispatchable;
@@ -118,6 +119,21 @@ public final class ClassFactory {
 
   public SClass[] getSuperclassAndMixins() {
     return superclassAndMixins;
+  }
+
+  /**
+   * This method is used to verify whether the class identified by `mixinId` was
+   * created from either the superclasses or any of other mixins encapsulated by
+   * this class factory.
+   */
+  public boolean isBasedOn(final MixinDefinitionId mixinId) {
+    VM.callerNeedsToBeOptimized("Result of isBasedOn should be cached");
+    for (SClass cls : getSuperclassAndMixins()) {
+      if (cls.isBasedOn(mixinId)) {
+        return true;
+      }
+    }
+    return false;
   }
 
   public void initializeClass(final SClass result) {
--- a/src/som/vmobjects/SClass.java
+++ b/src/som/vmobjects/SClass.java
@@ -173,7 +173,14 @@ public final class SClass extends SObjectWithClass {
     // assert instanceClassGroup != null || !ObjectSystem.isInitialized();
   }
 
-  private boolean isBasedOn(final MixinDefinitionId mixinId) {
+  /**
+   * This method checks whether a given class was created from `this`
+   * class. The check is made by comparing unique identifiers, the
+   * one given as an argument and the from the MixinDefintion encapsulated
+   * by this class. The identities match when the class identifier by the
+   * argument was created by this class.
+   */
+  public boolean isBasedOn(final MixinDefinitionId mixinId) {
     return this.mixinDef.getMixinId() == mixinId;
   }
 
@@ -194,7 +201,7 @@ public final class SClass extends SObjectWithClass {
     return superclass.isKindOf(clazz);
   }
 
-  public SClass getClassCorrespondingTo(final MixinDefinitionId mixinId) {
+  public SClass lookupClass(final MixinDefinitionId mixinId) {
     VM.callerNeedsToBeOptimized(
         "This should not be on the fast path, specialization/caching needed?");
     SClass cls = this;
@@ -204,8 +211,18 @@ public final class SClass extends SObjectWithClass {
     return cls;
   }
 
+  public SClass lookupClassWithMixinApplied(final MixinDefinitionId mixinId) {
+    VM.callerNeedsToBeOptimized(
+        "This should not be on the fast path, specialization/caching needed?");
+    SClass cls = this;
+    while (cls != null && !instanceClassGroup.isBasedOn(mixinId)) {
+      cls = cls.getSuperClass();
+    }
+    return cls;
+  }
+
   @ExplodeLoop
-  public SClass getClassCorrespondingTo(final int superclassIdx) {
+  public SClass lookupClass(final int superclassIdx) {
     SClass cls = this;
     for (int i = 0; i < superclassIdx && cls != null; i++) {
       cls = cls.getSuperClass();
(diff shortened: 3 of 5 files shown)


core-lib/TestSuite/MixinApplicationTests.ns
```diff
@@ -0,0 +1,54 @@
+(* Copyright (c) 2001-2017 see AUTHORS file
```
smarr: You don't like: https://github.com/smarr/SOMns/blob/master/core-lib/TestSuite/MixinTests.ns ?
Please place the tests next to the existing ones.
Please also check whether this resolves any of the issues with the currently deactivated tests in there.

src/som/interpreter/nodes/OuterObjectRead.java
```diff
@@ -87,6 +87,9 @@ public boolean isSuperSend() {
 
   protected final SClass getEnclosingClass(final SObjectWithClass rcvr) {
     SClass lexicalClass = rcvr.getSOMClass().getClassCorrespondingTo(mixinId);
```
smarr: This is silly nitpicking, but, I just had issues reading this code, so, could we use a `rcvrClass` local variable?
Then we also avoid twice the `.getSOMClass()`. Not that it matter performance-wise, but I have to think of one indirection less.

src/som/interpreter/objectstorage/ClassFactory.java
```diff
@@ -120,6 +121,15 @@ public boolean hasOnlyImmutableFields() {
     return superclassAndMixins;
   }
 
+  public boolean isBasedOn(final MixinDefinitionId mixinId) {
+    for (SClass cls : getSuperclassAndMixins()) {
```
smarr: Could you add a call to `VM.callerNeedsToBeOptimized("Result of isBasedOn should be cached")` before the loop?

src/som/interpreter/objectstorage/ClassFactory.java
```diff
@@ -120,6 +121,15 @@ public boolean hasOnlyImmutableFields() {
     return superclassAndMixins;
   }
 
+  public boolean isBasedOn(final MixinDefinitionId mixinId) {
```
smarr: Could you add a comment what the semantics of this now is?

src/som/vmobjects/SClass.java
```diff
@@ -173,7 +173,7 @@ public void initializeStructure(final MixinDefinition mixinDef,
     // assert instanceClassGroup != null || !ObjectSystem.isInitialized();
   }
 
-  private boolean isBasedOn(final MixinDefinitionId mixinId) {
+  public boolean isBasedOn(final MixinDefinitionId mixinId) {
```
smarr: Same here, could you add a comment what the semantics of this now is?
(not yet awake, and I haven't read this code in ages, so not sure what I was thinking here back then. sorry)

src/som/vmobjects/SClass.java
```diff
@@ -204,6 +204,16 @@ public SClass getClassCorrespondingTo(final MixinDefinitionId mixinId) {
     return cls;
   }
 
+  public SClass getClassWithMixinCorrespondingTo(final MixinDefinitionId mixinId) {
+    VM.callerNeedsToBeOptimized(
+        "This should not be on the fast path, specialization/caching needed?");
```
smarr: :) 👍

src/som/vmobjects/SClass.java
```diff
@@ -204,6 +204,16 @@ public SClass getClassCorrespondingTo(final MixinDefinitionId mixinId) {
     return cls;
   }
 
+  public SClass getClassWithMixinCorrespondingTo(final MixinDefinitionId mixinId) {
```
smarr: hmm, I don't like the name.
If I ask `clazz.getClass*()`, I think I expect the class of the class with some property.
How about just `getMixin(..)`?

Also see my comment for `getClassCorrespondingTo()`

src/som/vmobjects/SClass.java
```diff
@@ -204,6 +204,16 @@ public SClass getClassCorrespondingTo(final MixinDefinitionId mixinId) {
...
+        "This should not be on the fast path, specialization/caching needed?");
+    SClass cls = this;
+    while (cls != null && !instanceClassGroup.isBasedOn(mixinId)) {
+      cls = cls.getSuperClass();
+    }
+    return cls;
+  }
+
   @ExplodeLoop
   public SClass getClassCorrespondingTo(final int superclassIdx) {
```
smarr: And I suppose you were just following my example? Well, looks like i changed my mind here, didn't I?
Hmmm, well, don't know. The `CorrespondingTo` seems superfluous. Maybe it isn't.
Ok, I'll leave it up to you.

smarr (changes requested):
I have no idea whether the code is correct, haven't even tried to think about it. Sorry.

But, it looks good :)
Just some minor things.

Not sure what's going on with CI, we might need to update mx to get it working first. Would be worth a try.

smarr:
Looks like the CI issue was temporary.

core-lib/TestSuite/MixinApplicationTests.ns
ghost: I've examined the deactivated tests. Some success - two are fixed, three remain broken. Out of the three broken tests:

1. (testConflictingSlotsImmutable) I don't understand which answer is correct
2. (TestInitializer1) A slot (that should exist) is not found. Unsure how to debug
3. (TestInitializer3) I don't whether a class resulting from a mixin application be able to access mixed in methods annotated as private? 

Some more details below:


testConflictingSlotsImmutable - Broken

`c` is an instance of ConflictingSlotsImmutable, a class that inherits from Object and applies a mixin X that declares a public slot x. The initer method for ConflictingSlotsImmutable defines a new public slot x with initer expression `x = 1.`

After the initer method runs, `x` is equal to `1`. The test asserts that slot `x` is nil. I'm unsure which answer is correct.



testConflictingMethods - Fixed


`c` is an instance of ConflictingMethods, a class that inherits from `Y` and applies a mixin `Y2`. Both `Y` and `Y2` define a method `foo`. The test now passes with the correct result (`Y2`'s foo wins).


TestInitializer1 - Broken

`c` is an instance of `C`, which inherits from object and applies mixins X (first) and `X2` (second). Both `X` and `X2` define a public slot `x` with an initer expression. SOMns throws an assertion in this case (of which I'm unsure how to debug):

```
java.lang.AssertionError
    at som.vmobjects.SObject.getLocation(SObject.java:485)
    at som.vmobjects.SObject.writeSlot(SObject.java:509)
```

TestInitializer2 - Fixed

`c` is an instance of `C`, which inherits from object and applies mixins `X` (first) and `X2` (second). Both X and X2 define a public slot `x`. Both `X` and `X2` define an initer that assigns `x` a different number. The test now passes with the correct result (`X2`'s initer wins).



TestInitializer3 - Broken

`c` is an instance of `C`, which inherits from object and applies mixins `X` (first) and X2 (second). `X` defines a private method `cnt` and `X`'s initer method uses this variable to set a public field `x`. When the tests runs, `c` does not understand the message `cnt`. 

Should a class resulting from a mixin application be able to access mixed in methods annotated as private?

src/som/vmobjects/SClass.java
ghost: `getClassWithMixinCorrespondingTo` returns the receiver class or one of its superclasses (not the mixin's class). The class returned is that one who has had the given mixin applied. For example if we have an class `C` defined as:

```
Mixin = class()()
B = class Object <: Mixin ()()
C = class B ()()
```

Then the result of `C. getClassWithMixinCorrespondingTo(Mixin)` is `B`, because the mixin was applied to `B`. Does that make sense? Do you have a suggestion for the method name, given this example?

src/som/vmobjects/SClass.java
ghost: I agree that `CorrespondingTo` is not needed. I will remove. See my previous comment regarding the naming for `getClassWithMixinCorrespondingTo` and we we can go from there.

core-lib/TestSuite/MixinApplicationTests.ns
smarr: > Should a class resulting from a mixin application be able to access mixed in methods annotated as private?

This is a question for the spec. I don't know by hard, but I would suspect so.

About the slot related issues. I won't be able to dive into that at the moment, so, let's just ignore those for now. Except if you need that to work for Grace.

src/som/vmobjects/SClass.java
smarr: Wait, let's just talk about the java level and the naming, not the functionality on the language level.

The first thing that `getClassWithMixinCorrespondingTo(.)` can return is `this` (line 210).
The name of the method does not make me expect that.

If I ask any object in Java `this.getClass()`, I'll never get `this` returned, right?

smarr: Perhaps the get is bad. Perhaps what this is here is actually a lookup.
Perhaps just `lookupClass(mixinId)`?

core-lib/TestSuite/MixinApplicationTests.ns
ghost: Let's ignore it for now. The current implementation is at least a step in the right direction (at least, its sufficient for Grace since I won't use slots at the moment). I'd be happy to readdress the slot issues next month and make any changes to fix up three tests that remain broken.

ghost:
I've adapted the PR to meet the requested changes. I'm not sure that I've done a great job of the new docstrings (ClassFactory.isBasedOn, SClass.isBasedOn, and OuterObjectRead.getEnclosingClass). Please review before accepting!

smarr (approved):

smarr:
As you might have seen, I just briefly checked for the expected behavior with the Newspeak experts. Just to avoid reading the spec carefully, I suppose 😆 

I also compress the history a bit. Seemed useful to have the reason for the refactorings in the same commit. But, I probably have a different opinion on it every day...

Will merge once CI passed.

PR #190, opened 2017-10-11
Add support for Newspeak's Files

daumayr:
Adds support for file access similar to Newspeak, the major difference in functionality is the lack of streams.

- FileDescriptor used to read/write files, a integrated buffer is used to store read data and data to write.
- FilePath: represents a single file or directory and provides some operations (e.g. copy, move, size)
- FilePattern: represents a set of files or directories in the form of a Path with wildcards (*, ?) and provides operations (e.g. pathsDo: , delete).

```
f:: FilePath for: ('/directory/file'). //create a filepath.
d:: f open: #readwrite ifFail: [:r|]. //returns file descriptor.
d buffer at: 1 put: 127.
d write: 1 at: 0 ifFail: [:r|]. //write 01111111 at the files beginning.
```


smarr:
Thanks for the PR!

Haven't looked at the code yet. So, just some basic first questions:

The file gives us access to byte buffers, or what exactly is `buffer`? An integer array?
Any way to read stuff as a string?

Are you already using it for something?
Or do you plan to use it for something specific?

Would it be possible to use the provided changes to change [TestRunner](https://github.com/smarr/SOMns/blob/master/core-lib/TestSuite/TestRunner.ns#L85) to pick up tests automatically? At the moment, we are hardcoding them, which is of course not exactly ideal.

tests/files/test.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
smarr: Why did you chose to run those tests separately and not as unit tests with the other things in [TestRunner](https://github.com/smarr/SOMns/blob/master/core-lib/TestSuite/TestRunner.ns#L85)?

daumayr:
The buffer is a SOMns Array (long), reading stuff as a string is currently not supported, for that streams would be necessary.
Currently it is used in the Tests only, when expanding replay to deal with external communication I am going to make a benchmark that uses it.

I think it should be possible to pick up test automatically, streams would help

src/som/vmobjects/SFileDescriptor.java
```diff
@@ -0,0 +1,163 @@
...
+    try {
+      if (mode.getString().equals("read")) {
+        raf = new RandomAccessFile(f, "r");
+      } else {
+        raf = new RandomAccessFile(f, "rw");
+      }
+    } catch (FileNotFoundException e) {
+      fail.getMethod().invoke(new Object[] {fail, e.toString()});
+    }
+    /*
```
smarr: Please don't commit dead code

src/som/vmobjects/SFileDescriptor.java
```diff
@@ -0,0 +1,163 @@
...
+  private boolean open       = false;
+  private SArray  buffer;
+  private SSymbol mode;
+  private int     bufferSize = BUFFER_SIZE;
+
+  private RandomAccessFile raf;
+  private final File       f;
+
+  public static void setSOMClass(final SClass cls) {
+    // assert fileDescriptorClass == null || cls == null;
```
smarr: No dead code please

core-lib/Files.ns
```diff
@@ -0,0 +1,657 @@
+class Files usingVmMirror: vmMirror usingPlatform: platform = Value(
+(* This code was derived by converting the Strongtalk File classes to Newspeak, which is why the Sun Microsystems copyright and BSD license below applies.
```
smarr: Is this an unedited copy of the file?
If not, could you first have a commit with the original, and then the necessary changes?
Makes it much easier to track divergence.

Also, any chance you could improve the formatting of the file for human readers? (ideally in a separate commit)

core-lib/Files.ns
```diff
@@ -0,0 +1,657 @@
...
+    self moveAs: newPath
+      ifFail: [ :err <Symbol> |
+              self error: err	].
+    ^newPath
+  )
+
+  public simpleName ^ <String> = (
+    ^elements last name
+  )
+  (*
```
smarr: No dead code please (also in other places)

core-lib/Kernel.ns
```diff
@@ -314,6 +314,7 @@ class Kernel vmMirror: vmMirror = Object <: Value (
 
     (* Comparing *)
     public = argument = ( ^ vmMirror string: self equals: argument )
+    public ~= other = ( ^ (self = other) not )
```
smarr: Value defines the `<>`. Would be good if we could just use that and don't have to introduce the `~=`, which is meant for pointer objects only.

core-lib/Kernel.ns
```diff
@@ -651,6 +652,8 @@ class Kernel vmMirror: vmMirror = Object <: Value (
 
     public first = ( ^ isEmpty ifTrue: [nil] ifFalse: [storage at: firstIdx] )
 
+    public last = ( ^ isEmpty ifTrue: [nil] ifFalse: [storage at: (lastIdx - 1)] )
```
smarr: Please test this, where the other Vector things are tested

src/som/primitives/FilePrims.java
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  @GenerateNodeFactory
+  @Primitive(primitive = "file:write:at:ifFail:")
+  public abstract static class WriteFilePrim extends QuaternaryExpressionNode {
+    @Specialization
+    public final Object write(final SFileDescriptor file, final long nBytes,
+        final long offset, final SBlock fail) {
+      file.write((int) nBytes, offset, fail);
+      return Nil.nilObject;
```
smarr: I think it is common to return the written value.

src/som/primitives/FilePrims.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+  }
+
+  @GenerateNodeFactory
+  @Primitive(primitive = "file:openIfFail:")
+  public abstract static class FileOpenPrim extends BinaryExpressionNode {
+    @Specialization
+    public final Object fileOpen(final SFileDescriptor file, final SBlock handler) {
+      file.openFile(handler);
+      return Nil.nilObject;
```
smarr: Might be nicer to return the `file`?

src/som/primitives/FilePrims.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+  }
+
+  @GenerateNodeFactory
+  @Primitive(primitive = "file:setMode:")
+  public abstract static class FileSetModePrim extends BinaryExpressionNode {
+    @Specialization
+    public final Object setMode(final SFileDescriptor file, final SSymbol mode) {
+      file.setMode(mode);
+      return Nil.nilObject;
```
smarr: Might be nicer to return `file`?

src/som/primitives/FilePrims.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+  }
+
+  @GenerateNodeFactory
+  @Primitive(primitive = "file:setBufferSize:")
+  public abstract static class FileSetBufferSizePrim extends BinaryExpressionNode {
+    @Specialization
+    public final Object setBufferSize(final SFileDescriptor file, final int size) {
+      file.setBufferSize(size);
+      return Nil.nilObject;
```
smarr: Might be nicer to return `file`?

src/som/primitives/FilePrims.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+  }
+
+  @GenerateNodeFactory
+  @Primitive(primitive = "fileClose:")
+  public abstract static class CloseFilePrim extends UnaryExpressionNode {
+    @Specialization
+    public final Object closeFile(final SFileDescriptor file) {
+      file.closeFile();
+      return Nil.nilObject;
```
smarr: Might be nicer to return `file`?

src/som/primitives/PathPrims.java
```diff
@@ -0,0 +1,183 @@
...
+public final class PathPrims {
+  @GenerateNodeFactory
+  @Primitive(primitive = "pathContents:")
+  public abstract static class PathContentsPrim extends UnaryExpressionNode {
+    @Specialization
+    public final Object getContents(final String directory) {
+      File f = new File(directory);
+      String[] content = f.list();
+      Object[] o = new Object[content.length];
+      for (int i = 0; i < content.length; i++) {
```
smarr: Is this just `System.arraycopy(...)`?

src/som/primitives/PathPrims.java
```diff
@@ -0,0 +1,183 @@
...
+        o[i] = content[i];
+      }
+
+      if (content != null) {
```
smarr: shouldn't this be checked before you do `content.length` in line 33?

src/som/primitives/PathPrims.java
```diff
@@ -0,0 +1,183 @@
...
+
+  @GenerateNodeFactory
+  @Primitive(primitive = "path:copyAs:ifFail:")
+  public abstract static class FileCopyPrim extends TernaryExpressionNode {
+    @Specialization
+    public final Object copyAs(final String source, final String dest, final SBlock fail) {
+      try {
+        Files.copy(Paths.get(source), Paths.get(dest), new CopyOption[0]);
+      } catch (IOException e) {
+        fail.getMethod().invoke(new Object[] {fail, e.toString()});
```
smarr: would be nicer if you'd use a block dispatch node here instead of doing it manually, and unoptimized.

src/som/primitives/PathPrims.java
```diff
@@ -0,0 +1,183 @@
...
+
+  @GenerateNodeFactory
+  @Primitive(primitive = "pathCreateDirectory:ifFail:")
+  public abstract static class CreateDireytoryPrim extends BinaryExpressionNode {
+    @Specialization
+    public final Object createDirectory(final String dir, final SBlock fail) {
+      try {
+        Files.createDirectories(Paths.get(dir), new FileAttribute<?>[0]);
+      } catch (IOException e) {
+        fail.getMethod().invoke(new Object[] {fail, e.toString()});
```
smarr: block dispatch node for all these block things?

src/som/primitives/PathPrims.java
```diff
@@ -0,0 +1,183 @@
...
+      }
+      return Nil.nilObject;
```
smarr: I don't like all those nil returns, usually, self is returned implicitly in Smalltalk/Newspeak

src/som/primitives/PathPrims.java
```diff
@@ -0,0 +1,183 @@
...
+
+  @GenerateNodeFactory
+  @Primitive(primitive = "pathLastModified:")
+  public abstract static class LastModifiedPrim extends UnaryExpressionNode {
+    @Specialization
+    public final String lastModified(final String dir) {
+      try {
+        return Files.getLastModifiedTime(Paths.get(dir), new LinkOption[0]).toString();
+      } catch (IOException e) {
+        throw new RuntimeException(e);
```
smarr: This isn't a good solution. Would be nicer if you throw a proper Smalltalk exception here.
NotAValue, or index out of bounds are examples, I think

src/som/vmobjects/SFileDescriptor.java
```diff
@@ -0,0 +1,163 @@
...
+    buffer = new SMutableArray(storage, Classes.arrayClass);
+
+    try {
+      if (mode.getString().equals("read")) {
+        raf = new RandomAccessFile(f, "r");
+      } else {
+        raf = new RandomAccessFile(f, "rw");
+      }
+    } catch (FileNotFoundException e) {
+      fail.getMethod().invoke(new Object[] {fail, e.toString()});
```
smarr: It would be better to use a block dispatch node, which you probably need to pass in as a parameter from the AST node where you call this method.

src/som/vmobjects/SFileDescriptor.java
```diff
@@ -0,0 +1,163 @@
...
+    super(fileDescriptorClass, fileDescriptorClass.getInstanceFactory());
+    f = new File(uri);
+  }
+
+  public void openFile(final SBlock fail) {
+    long[] storage = new long[bufferSize];
+    buffer = new SMutableArray(storage, Classes.arrayClass);
+
+    try {
+      if (mode.getString().equals("read")) {
```
smarr: Would be nicer if you just obtain the symbol for `"read"` and then compare identity

src/som/vmobjects/SFileDescriptor.java
```diff
@@ -0,0 +1,163 @@
...
+        raf = new RandomAccessFile(f, "r");
+      } else {
```
smarr: Also, what does Newspeak do if you pass in `#foobar`? or `nil`, or whatever for `mode`? here we get crashes, or if it is any symbol, we get write access. Might be better to have an exception or call the failure block?

src/som/vmobjects/SFileDescriptor.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+    /*
+     * open with write only
+     * raf.getChannel().truncate(0);
+     */
+
+    open = true;
+  }
+
+  public void closeFile() {
```
smarr, line 58: Hm, you got java doc on the next method, but not here?

src/som/vmobjects/SFileDescriptor.java
```diff
@@ -0,0 +1,163 @@
...
+    for (int i = 0; i < bufferSize; i++) {
+      storage[i] = buff[i];
+    }
+
+    return bytes;
+  }
+
+  public void write(final int nBytes, final long position, final SBlock fail) {
+    if (!open) {
+      fail.getMethod().invoke(new Object[] {fail, "File not open"});
```
smarr: block node please, also the other cases.

src/som/vmobjects/SFileDescriptor.java
```diff
@@ -0,0 +1,163 @@
...
+      return;
+    }
+
+    long[] storage = (long[]) buffer.getStoragePlain();
+    byte[] buff = new byte[bufferSize];
+
+    for (int i = 0; i < bufferSize; i++) {
+      // punish users for using this loophole
```
smarr: ehm, no?! Throw a proper language level exception, please.

src/som/vmobjects/SFileDescriptor.java
```diff
@@ -0,0 +1,163 @@
...
+
+    open = true;
+  }
+
+  public void closeFile() {
+    try {
+      raf.close();
+      open = false;
+    } catch (IOException e) {
+      e.printStackTrace();
```
smarr: A Smalltalk exception would be much nicer

smarr (changes requested):
There are a couple of things, I'd prefer to see changed I think:

- return values, self/this is often preferable to nil
- block dispatch nodes would be better than hardcoded invokes
- please remove dead code
- tests better integrated with unit tests
and a few other things

core-lib/Files.ns
daumayr: The File API was split into multiple files and classes which I merged, so I don't think committing changes on top of the newspeak original would make sense. Also the formatting was worse. I am going to change that comment to say it's based on NewSpeak File classes.

core-lib/Files.ns
daumayr: it's a public method accessible by users (which isn't used internally)

core-lib/Kernel.ns
daumayr: ok, i didn't know about  the <> operator

src/som/primitives/PathPrims.java
daumayr: the primitive doesn't have a self, or should i return the string argument?

core-lib/Files.ns
smarr: `writeStream` looks like it is in a comment, so, it's not accessible?

core-lib/Files.ns
smarr: Then please carefully attribute the source.
We need to be able to at least find the original files, ideally with an URL into the original repo for each, including version information.

src/som/primitives/PathPrims.java
smarr: Hm, ok, for these cases `nil` might be fine. In other case, the first parameter might be more useful.

core-lib/Files.ns
daumayr: I misinterpreted the comment, only the (* of the comment was visible. I decided to keep the stream based methods as comments for the case that streams are implemented in the future. Those methods could be useful for reading text from files.

tests/files/test.sh
daumayr: hm, right, i didn't end up needing batch magic for cleanup

core-lib/Kernel.ns
daumayr: i am going to add a test for `first` aswell  ;)

smarr:
Ok, I did some more polishing, moving test files a bit around, fix formatting a little. Nothing much of consequence.
I squashed the changes, too, since the history didn't reflect anything useful.

@daumayr if you don't have any further things, I'll merge it.

smarr:
@richard-roberts ah, perhaps interesting for you to have a brief look. This adds file support based on Newspeak. Might be useful for Grace.

daumayr:
@smarr you can merge it, the broken tests are already broken in dev, i'll take a look at that separately.

ghost:
Looks nice, thanks @daumayr. Early next year I hope to get our self-hosting Grace compiler running on SOMns, these changes will be useful for getting that going!

PR #191, opened 2017-10-15
Exit VM on uncaught exception

smarr:
Request a VM exit when we catch an uncaught exception.

Was there any reason we didn't do this before?
Not sure why this isn't done.
This leaves the VM hanging on error, which is problematic on the CI systems, and strange locally.

@daumayr could you have a look at this?


daumayr:
Looks fine, the failed CI for the second replay test has probably a different cause (sometimes the trace can't be replayed)

PR #192, opened 2017-10-16
Add tool to identify candidates for super-instructions

smarr:
This PR adds a new Truffle tool to instrument all nodes, track their execution, and determine node configurations that are frequently used together.
These node configurations might benefit from being converted into a single node that reduces the interpretation overhead.

This is work by @fredreichbier.

@fredreichbier, could you please review the PR, and check whether everything looks ok for you?

This is a first version, I might still do some more cosmetic changes.


src/tools/dym/superinstructions/CandidateDetector.java
```diff
@@ -0,0 +1,164 @@
...
+   */
+  public String detect() {
+    // Find all activation contexts with traces of length n = 2,
+    // exclude traces containing SequenceNodes (they do not yield
+    // viable candidates for super-instructions),
+    // sort them by activation counts in descending order,
+    // fetch the top CONSIDER_TOP_CONTEXTS activation contexts
+    // and construct a candidate for each of them.
+    List<ActivationContext> sorted =
+        contexts.keySet().stream().filter(context -> context.getNumberOfClasses() == 3)
```
smarr: @fredreichbier here I was wondering why that's a three. The comment above talks about 2, I believe.

fredreichbier: Oops you're right, there is an inconsistency here.
Given a trace `[C_0, i_0, ... C_n]`, `getNumberOfClasses` does not actually return `n`, it returns `n+1` (because e.g. a trace of length 1 is `[C_0, i_0, C_1]`, so `n=1`, but `getNumberOfClasses() == 2`). I guess the most sensible way to resolve this would be to modify the documentation of `getNumberOfClasses`.

smarr:
This PR was blocked by failing tests on Linux.
Turns out that the final step to create the report was contain some non-deterministic bits.
The number of results was limited to 100 before sorting them in a fully deterministic way. Sorting relied only on the score, which means order for same score is not defined.

Cutting off at that point could result in slightly different reports being generated.
Now, we make sure all candidates are unique, sort them by score as well as their other attributes, and then limit the result set to 100 elements as the last step. This leads the tests being sufficiently robust.

PR #194, opened 2017-11-22
Update Truffle to state of 2017-11-09

smarr:
Update Truffle and Graal to version of 2017-11-09.
I selected that version because afterwards, deprecated APIs where removed.
This is a first step. The next step will be to update to the latest version.


smarr:
The update was done in https://github.com/smarr/SOMns/commit/72cd8c95e5cd5894c05d7e365965b35df472c34f

PR #201, opened 2017-12-14
Fix mixin initializers

daumayr:
This PR fixes issues with the lookup of mixin initialisers of mixins used in superclasses. 
Previously the following example would not call the initializer of Basic, resulting in getField returning nil. 

```
class Test usingPlatform: platform = Value ()(
  class Basic = (
    | aField ::= false. |
    'basic' println.
    )
  (
    getField = ( ^aField )
  )

  class Super = Object <: Basic (
    'super' println.
  )()

  class Mixin = ()()

  class Sub = Super <: Mixin (
    'Sub' println.
  )(
    public aMethod = (
      getField println.
    )
  )

  public main: args = (
    |s|
    s:: Sub new.
    s aMethod.
    ^ 0
  )
)
```

The mixin initializer were Protected, causing a lookupPrivate call to default to lookupMethod, as the mixin initialisers of Super and Sub have identical names, lookupMethod returns the initializer for Mixin which is found at the beginning of the lookup in the class Sub

The problem was solved by making the mixin initialisers private.

### DIFF
--- a/src/som/compiler/MixinDefinition.java
+++ b/src/som/compiler/MixinDefinition.java
@@ -445,7 +445,7 @@ public final class MixinDefinition {
 
     return initializerBuilder.splitBodyAndAssembleInitializerAs(
         MixinBuilder.getInitializerName(primaryFactoryName, mixinId),
-        body, AccessModifier.PROTECTED, initializerSource);
+        body, AccessModifier.PRIVATE, initializerSource);
   }
 
   private void addSlots(final HashSet<SlotDefinition> instanceSlots,
--- a/core-lib/TestSuite/MixinTests.ns
+++ b/core-lib/TestSuite/MixinTests.ns
@@ -203,6 +203,45 @@ class LanguageTests usingPlatform: platform testFramework: minitest = (
       )
     )
 
+    class InitMixinFields = ()(
+      class MixinWithFields = (| public x ::= 23. public y = 24. |)
+      ( 
+        public getx = ( ^ x )
+
+        public gety = ( ^ y )
+      )
+
+      class SuperClassWithMixin = Object <: MixinWithFields (| public z ::= 25. |)(
+        public getz =( ^ z )
+      )
+
+      class AnotherMixin = (| u ::= 0. |)( public getu = (^ u) )
+
+      public class SubClassWithMixin = SuperClassWithMixin <: AnotherMixin ()()
+
+      public class SubClassWithoutMixin = SuperClassWithMixin ()()
+    )
+
+    (*Temporary duplicate until issue #202 is fixed*)
+    class InitMixinFields2 = ()(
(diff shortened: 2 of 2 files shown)


src/som/compiler/MixinBuilder.java
```diff
@@ -607,8 +611,10 @@ public static SSymbol getInitializerName(final SSymbol selector) {
   }
 
   public static SSymbol getInitializerName(final SSymbol selector,
-      final int mixinId) {
-    return symbolFor("initializer`" + mixinId + "`" + selector.getString());
+      final int mixinId, final MixinDefinitionId mixinDefinitionId) {
+    return symbolFor(
+        "initializer`" + mixinId + "`" + "_" + mixinDefinitionId + "_"
```
smarr: Do we really need both, `mixinId` and `mixinDefinitionId`?

src/som/compiler/MixinDefinition.java
```diff
@@ -431,7 +431,8 @@ private void reportErrorAndExit(final String msgPart1, final String msgPart2) {
                       .errorExit(line + msgPart1 + name.getString() + msgPart2);
   }
 
-  private SInitializer assembleMixinInitializer(final int mixinId) {
+  private SInitializer assembleMixinInitializer(final int mixinId,
```
smarr: `mixinId` still needed?

src/som/compiler/MixinDefinition.java
```diff
@@ -367,7 +367,7 @@ protected boolean hasOnlyImmutableFields(final HashSet<SlotDefinition> instanceS
           }
         }
 
-        SInitializer mixinInit = cdef.assembleMixinInitializer(i);
```
smarr: I suspect the for loop above can be changed to be a for each loop over `mixins`, we don't need the `i` anymore,  I think.

smarr (changes requested):
Great, looks like a clean and simple fix! :)

And might allow for the mentioned simplifications.

Could you also please add tests to:

https://github.com/smarr/SOMns/blob/master/core-lib/TestSuite/MixinTests.ns

Would be good to test a couple of things.
Deeper chains and mixin in the same thing at various levels, and various orders. Ideally tests that fail before applying the fix.

src/som/compiler/MixinBuilder.java
daumayr: mixinId is needed to distinguish multiple mixins used by the same class

smarr: Ah, and please stick with a single style of separators.
For some reason, I decided that the back tick is good, probably, because it can't be part of a normal identifier while `_` can.

smarr (approved):

smarr:
Reworded the commit message a bit to state the problem.

Thanks for debugging and fixing this!

PR #204, opened 2017-12-17
Adopt Streams Library from Newspeak

daumayr:
The Streams library of Newspeak was adapted to SOMns, it provides streams to read/write from files, collections (Vectors), and also offers character conversion. 
The stream related file methods were added to the file library, making it easier to read/write files, as it is no longer necessary to directly manipulate the buffers.

Below is an example that shows how easy an entire text file can be read:
```
fp:: FilePath for: '/home/user/aFile'.
cis:: fp charInputStream.
text:: cis next: (cis size).
cis close.
```

Similar for writing:
```
fp:: FilePath for: '/home/user/aFile'.
cos:: fp charOutputStream.
cos putAll: 'aString'.
cos close.
```


daumayr:
The contents of this PR are based on PR #201

core-lib/Files.ns
```diff
@@ -357,6 +363,12 @@ DAMAGE.>> *)
     public pattern: p <String> = (
       pattern_0:: p
     )
+
+    (*stream stuff*)
```
smarr: Not a helpful comment, please drop...

core-lib/Files.ns
```diff
@@ -548,6 +560,47 @@ DAMAGE.>> *)
     public simpleName ^ <String> = (
       ^ elements last name
     )
+
+    (*stream stuff*)
```
smarr: Not a helpful comment, please drop...

core-lib/Files.ns
```diff
@@ -548,6 +560,47 @@ DAMAGE.>> *)
...
+    charInputStream ^ <CharInputStream> = (
```
smarr: that's not api? that's implicitly protected? same with following methods?

core-lib/Kernel.ns
```diff
@@ -71,6 +71,9 @@ class Kernel vmMirror: vmMirror = Object <: Value (
     public println  = ( vmMirror printNewline: self asString )
     public halt     = ( vmMirror halt: self )
     public asString = ( ^ 'instance of ' + (vmMirror objClassName: self) )
+    public printOn: stream = (
```
smarr: That's debugging functionality?

core-lib/Kernel.ns
```diff
@@ -410,8 +413,18 @@ class Kernel vmMirror: vmMirror = Object <: Value (
       ^ self substringFrom: argument to: argument
     )
 
+    public do: block = (
```
smarr: Why do we need #do: on strings?
Also, could you please add a test for any addition to Kernel.ns?

core-lib/Kernel.ns
```diff
@@ -410,8 +413,18 @@ class Kernel vmMirror: vmMirror = Object <: Value (
...
+      1 to: self length do: [ :i |
+        block value: (self charAt: i).
+      ]
+    )
+
     (* Printing *)
     public print = ( vmMirror printString: self )
+  ) : (
+    public from: charArray = (
```
smarr: What's the error when `charArray` isn't what it is supposed to be? Invalid arg exception or something alike? would be good to document.

core-lib/Kernel.ns
```diff
@@ -424,6 +437,46 @@ class Kernel vmMirror: vmMirror = Object <: Value (
     public print    = ( '#' print. super print )
   )
 
+  public class Character = String ()
+  ( 
+  ) : (
+    public asciiValue: val = (
```
smarr: What's this? an empty method?

core-lib/Kernel.ns
```diff
@@ -424,6 +437,46 @@ class Kernel vmMirror: vmMirror = Object <: Value (
...
+
+    )
+    
+    (*convert to char*)
+    public fromValue: val = (
```
smarr: isn't `from:` sufficient? and the `val` should perhaps be `anInt`? or document the type, please

core-lib/Kernel.ns
```diff
@@ -424,6 +437,46 @@ class Kernel vmMirror: vmMirror = Object <: Value (
...
+      ^vmMirror charFrom: val
+    )
+
+    public value: char = (
```
smarr: No idea what this is supposed to do. What does it return? an integer? What kind of value? ASCII, utf-8?

core-lib/Kernel.ns
```diff
@@ -424,6 +437,46 @@ class Kernel vmMirror: vmMirror = Object <: Value (
...
+      ^vmMirror charValue: char
+    )
+
+    (*constants*)
+    public backspace = (^'\b')
```
smarr: Why are these needed?

core-lib/Kernel.ns
```diff
@@ -424,6 +437,46 @@ class Kernel vmMirror: vmMirror = Object <: Value (
...
+    public ff = (^'\f')
+    public lf = (^'\n')(*)
+    public newPage = ()
+    public nl = ()*)
+    public nul = (^'\0')
+    public space = (^'\s')
+    public tab = (^'\t')
+  )
+
+  public class StringBuilder = Object (
```
smarr: Is this needed? At least remove the unnecessary whitespace:

```
public class StringBuilder = Object ()() : (
  public new: size = ( ^ vmMirror array: self new: size )
)
```

smarr (changes requested):
Thanks a lot for preparing this. Here the first part of review.

src/som/primitives/StringPrims.java
```diff
@@ -153,4 +155,41 @@ public final String doString(final String receiver, final long idx) {
...
+    }
+  }
+
+  @GenerateNodeFactory
+  @Primitive(primitive = "charFrom:")
+  public abstract static class CharFromPrim extends UnaryExpressionNode {
+
+    @Specialization
+    public final String doString(final long val) {
+      return new String(new char[] {(char) val});
```
smarr: This is very forgiving, not sure we want this, might be surprising. Might need to check the val to be in range, and throw an exception otherwise?

src/som/primitives/StringPrims.java
```diff
@@ -153,4 +155,41 @@ public final String doString(final String receiver, final long idx) {
...
+  @GenerateNodeFactory
+  @Primitive(primitive = "stringFrom:")
+  public abstract static class StringFromPrim extends UnaryExpressionNode {
+
+    @Specialization
+    public final String doString(final SArray chars) {
+      Object[] storage = chars.getObjectStorage(chars.ObjectStorageType);
+      StringBuilder sb = new StringBuilder(storage.length);
+      for (Object o : storage) {
+        sb.append((String) o);
```
smarr: This doesn't ensure that the strings are all 1-char long, right? Might need to error on invalid 'chars'?

src/som/primitives/StringPrims.java
```diff
@@ -153,4 +155,41 @@ public final String doString(final String receiver, final long idx) {
...
+    }
+  }
+
+  @GenerateNodeFactory
+  @Primitive(primitive = "charValue:")
+  public abstract static class CharValuePrim extends UnaryExpressionNode {
+
+    @Specialization
+    public final long doString(final String c) {
+      assert c != null && c.length() == 1;
```
smarr: This should probably be throwing a SOMns exception instead of failing an assertion.

src/som/vmobjects/SFileDescriptor.java
```diff
@@ -41,13 +41,13 @@ public void openFile(final SBlock fail, final BlockDispatchNode dispatchHandler)
...
-      if (mode == Symbols.symbolFor("read")) {
-        raf = new RandomAccessFile(f, "r");
-      } else if (mode == Symbols.symbolFor("readwrite")) {
-        raf = new RandomAccessFile(f, "rw");
-      } else {
-        dispatchHandler.executeDispatch(new Object[] {fail, "invalid access mode"});
-      }
+      this.access = AccessModes.valueOf(mode.getString());
+    } catch (Exception e) {
+      dispatchHandler.executeDispatch(new Object[] {fail, "invalid access mode"});
```
smarr: Should the error include the wrong value, `mode`? Would make it easier to debug.

src/som/vmobjects/SFileDescriptor.java
```diff
@@ -157,4 +167,17 @@ public void setBufferSize(final int bufferSize) {
   public void setMode(final SSymbol mode) {
     this.mode = mode;
   }
+
+  private enum AccessModes {
+    read("r"), write("rw"), readwrite("rw");
```
smarr: I think it is common to have enum members be named all caps, since they are constants.

tests/streams/.gitignore
```diff
@@ -0,0 +1 @@
+*.txt
```
smarr: is this needed?

smarr (changes requested):
Another bunch of things.
For the whole 'exception' and strictness discussion, if you prefer not to do that now, I am fine with skipping it for now, but please create an issue to document it as an open TODO.

tests/streams/.gitignore
daumayr: It prevents you from accidentally committing files created by tests

smarr: Could you just add this in the top-level .gitignore that already exists? (probably as `tests/streams/*.txt`)

core-lib/Files.ns
```diff
@@ -548,6 +554,36 @@ DAMAGE.>> *)
...
+    public charInputStream ^ <CharInputStream> = (
+      ^CharacterReadConverter on: self readStream.
+    )
+
+    public charOutputStream ^ <CharOutputStream> = (
+      ^CharacterInputOutputConverter on: self writeStream
+    )
+
+    public inputStream ^ <InputStream[Integer]> = (
+      (* We are using this for now since read streams are faster at things like upTo: than input streams *)
```
smarr: Is this your comment, a preexisting one?
Since this is a performance issue, please mark origin (if you, just prefix by `DA:` for instance, and possibly date) so that we know in the future.

core-lib/Files.ns
```diff
@@ -548,6 +554,36 @@ DAMAGE.>> *)
...
+      ^self readStream
+    )
+
+    public outputStream ^ <OutputStream[Integer]> = (
+      ^ExternalOutputStream onDescriptor: (self open: #write ifFail: [ :err <ExternalError> | self error: err ])
```
smarr: What's `ExternalError`?
Is this an exception?
If so, perhaps it is better to just signal (i.e. throw it here) instead of using `self error:` which probably quits the VM.

core-lib/Files.ns
```diff
@@ -548,6 +554,36 @@ DAMAGE.>> *)
...
+    )
+
+    public readStream ^ <ExternalReadStream> = (
+      ^ExternalReadStream onDescriptor:
+      (self open: #read ifFail: [ :err <ExternalError> | self error: err ])
```
smarr: same comment as before, and also indentation looks strange.

core-lib/Files.ns
```diff
@@ -548,6 +554,36 @@ DAMAGE.>> *)
...
+    )
+
+    public readWriteStream ^ <ExternalReadWriteStream> = (
+      ^ExternalReadWriteStream onDescriptor: (self open: #readwrite ifFail: [ :err <ExternalError> | self error: err ])
```
smarr: Same as above, also for following method.

core-lib/Kernel.ns
```diff
@@ -895,6 +911,16 @@ class Kernel vmMirror: vmMirror = Object <: Value (
     )
   )
 
+  public class InvalidArgument signalWith: argument = Exception (
```
smarr: Hm, is the name from Newspeak?
I think, I'd like the class name to be explicit about being an exception.
Possibly named `InvalidArgumentException`, or `ArgumentException` if you are concerned about length?

For the naming, we should try to pick something we'd like to follow as guide, something to follow when it isn't in Newspeak already.
I checked
- Java: IllegalArgumentException
- C#: ArgumentException
- Dart: ArgumentError

any preferences? Perhaps Dart is the most modern one, might be a good source for inspiration?

core-lib/Kernel.ns
```diff
@@ -928,4 +954,8 @@ class Kernel vmMirror: vmMirror = Object <: Value (
   private signalTOCannotBeValues: aClass = (
     TransferObjectsCannotBeValues signalWith: aClass
   )
+
+  private signalInvalidArgument: message = (
+    InvalidArgument signalWith: message.
```
smarr: We should really build a nice exception thrower in Java, to avoid these unnecessary helpers...

core-lib/Streams.ns
```diff
@@ -0,0 +1,1037 @@
...
+			^false
+		)
+
+		upTo: delimiter <Object> ^ <Vector[E]> = (
+			| s <SeqCltn[E]> |
+			s:: self accumulateUntil: [ self atEndOrPeekIs: delimiter ].
+			self nextIfAbsent: [ ].
+			^s
+		)
+	) : (
```
smarr, line 172: Most of the methods above are protected, is that intentional? Only few are public interface, if I see it correctly.

core-lib/Streams.ns
```diff
@@ -0,0 +1,1037 @@
...
+	)
+
+	public class BasicOutputStream = (
+	(* This class provides the standard implementation of the OutputStream protocol.  It implements all functionality in terms of the message #nextPut:, which subclasses must provide.
```
smarr: I'd like class comments to precede the class normally, I think.
Seems pretty standard in most languages?

core-lib/Streams.ns
```diff
@@ -0,0 +1,1037 @@
...
+			^con isString
+				ifTrue: [ con ]
+				ifFalse: [ String from: (con asArray) ]
+		)
+	) : (
+	)
+
+	public class CharacterInputConverter = CharacterInput (
+	(* Instances are used to create a CharInputStream that wraps a byte input stream*)
+	| binaryStream <STRM> |) (
```
smarr: STRM?

core-lib/Streams.ns
```diff
@@ -0,0 +1,1037 @@
...
+
+		public size = (
+			^binaryStream size.
+		)
+
+		public initBinaryStream: binaryStrm <STRM> = (
+			self binaryStream: binaryStrm
+		)
+
+		upTo: del <Object> ^ <Str> = (
```
smarr: Str?

smarr: I think we want a basic name check for type annotations.
To make sure that we at least use things that exist :D

core-lib/Streams.ns
```diff
@@ -0,0 +1,1037 @@
...
+			| bytes <Vector[Integer]> s <String> lfs <Integer> j <Integer> delimiter <Character> |
```
smarr: could you add a line break after the types?
This feels hard to read :(

core-lib/Streams.ns
```diff
@@ -0,0 +1,1037 @@
...
+		onDescriptor: d <DESCRIPTOR> ^ <Instance> = (
+				^self new initDescriptor: d
+		)
+	)
+
+	usingCrLfConvention = (
+		(* The current code only distinguishes systems that use CR and systems use CRLF for line endings. In order to generate files that are printed correctly on systems that use LF (Linux, OS X), we default to CRLF for now. *)
+		^ true
+	)
+) : (
```
smarr, line 1183: This is unfortunately really to much for me to review.
I don't have a good understanding of what's going on, and what the overall design is :(

core-lib/Streams.ns
```diff
@@ -0,0 +1,1037 @@
...
+			self nextBufferPosition: 0.
+		)
+
+		postCopy = (
+			super postCopy.
+			self descriptor: self descriptor copy.
+			self initBuffer.
+		)
+	) : (
+		onDescriptor: d <DESCRIPTOR> ^ <Instance> = (
```
smarr: DESCRIPTOR?
Instance?
Is this method needed?
Seems to be protected.

core-lib/TestSuite/StreamTests.ns
```diff
@@ -0,0 +1,126 @@
...
+      1 to: 20 do: [ :i|
+        assert: (t at: i) = i.
+      ].
+
+      assert: cc atEnd.
+      
+      cc close.
+    )
+
+    public testReadBinary2 = (
```
smarr: What's the difference between the two tests?
What is their purpose?
I can't really tell from the method names.
And the assertions aren't self explanatory either.

src/som/vm/constants/KernelObj.java
```diff
@@ -33,6 +33,17 @@ public static Object signalException(final String selector, final Object receive
     return disp.invoke(new Object[] {KernelObj.kernel, rcvr.getSOMClass()});
   }
 
+  public static Object signalException(final String selector, final String receiver) {
```
smarr: Hm, I don't like the code duplication with the method above, and the name doesn't isn't really telling me why there are two methods. They do slightly different things, but perhaps the first version is kind of wrong.

Also, would be good to clean up all the duplication around exception throwing. There are quite a number of uses of lookupPrivate, which indicate the same pattern (PathPrims, ArraySetAllStrategy, ...)

smarr (changes requested):
Thanks for the update!

There are various little things, any opinion on those?

smarr:
For the failing test, the super-instruction candidate tool works like the dym tool. `./tests/superinstructions/test.sh update` should do the trick.

core-lib/Kernel.ns
daumayr: yes, would make introduction of new Exceptions easier

core-lib/Streams.ns
daumayr: I removed some dead code and made some more public, the original Newspeak implementation is not well designed in that regard...
Once the Streams are used more (e.g. for the Newspeak parser), we will know what needs to be changed/added.

smarr:
@daumayr I fixed the incomplete author info on some commits. Hope you don't mind.

And, I'll try to break this up in a couple different PRs. There is just too much going on here for me to make sense of.
First, I'll split out the character stuff.
And, the file stuff probably, too.

smarr (approved):
Ok, let's merge this and improve in place later.
Still don't have a great understanding of how things work, but will probably get polished step by step once merged.

PR #209, opened 2017-12-28
Add fork/join benchmarks

smarr:
This PR adds a few benchmarks based on the [Work–Stealing Without The Baggage](http://users.cecs.anu.edu.au/~steveb/downloads/pdf/ws-oopsla-2012.pdf) paper.

The original version originates from: https://github.com/vivkumar/TryCatchWS/tree/master/test
But was first ported to Java's fork/join framework here:
https://github.com/smarr/are-we-fast-yet/tree/fork-join/benchmarks/Java/src/forkjoin

The list of ported benchmarks includes:
 - CilkSort
 - Fib(onacci)
 - Integrate
 - Jacobi
 - LUDecomposition
 - NQueens
 - QuickSort

It does not yet include FFT and UTS.
FFT is just too much code for the moment.
And UTS does some bit shifting, which is not yet nicely supported in SOMns.

The benchmarks are generally not very sophisticated, even in their optimized version, they rarely achieve good scalability on Java's fork/join framework. Most do not use a sequential cut off, and thus task creation overhead is rather high.

To assess the characteristics of the benchmarks, they are all available in four versions:
 - Seq: no spawns, everything is done sequentially
 - Opt: uses spawns, but does at least one task locally
 - Nai: uses spawns for all tasks, so, it is only joining locally
 - One: spawns a task, and waits for it to finish. Thus, this is the worst case scenario, introducing f/j overhead without any gain.

Thanks to @VAISHALI-DHANOA, who started this work.

@eleinadani @eregon, this might be interesting to you. If we continue this, it might be interesting to port to Ruby. The benchmarks I got are pretty straight forward to translate, one just needs a lightweight fork/join mechanism. And, of course, pretty heavy on array data structures.

@daumayr possibly interesting for you, another bunch of benchmarks, and a change to the harness to add support for running benchmarks by file name.


eregon:
Nice set of benchmarks!
Would definitely be interested to port some to Ruby and see how they run and what we can optimize.

Issue #94, opened 2017-01-18
Add support for blocks with arbitrary number of arguments


smarr:
Currently, we only support blocks with up to two arguments (in addition to self).
So, the best we can do is `[:a :b | #foo ] value: 1 with: 3`.

Thus, we can't have blocks with 3 or more arguments. The implementation is currently also based on Smalltalk classes `Block1`, `Block2`, `Block3` with the corresponding interface.

It would be good to lift that restriction and enable blocks with more arguments.
I am not sure what the best way forward is.
I am also not sure whether blocks should support currying or 'culling' (I think that's the term used in Pharo). So, basically the question is whether a block should handle activation with insufficient amount of arguments.

We should look into Newspeak, Pharo, TruffleMate, Grace, etc to see options on how to handle this.

smarr:
A solution should also resolve the problem with issue #21.

smarr:
Think @charig's solution is here: https://github.com/charig/TruffleMATE/pull/31

smarr:
@richard-roberts Did you see #215? @daumayr already implemented a solution. I just forgot about it. Is this what you need? Will try to look into merging ASAP.

ghost:
@smarr This is what would I would have done. It would be good to get this into the dev branch if you have time (no rush though, it's only causing one Grace test to fail for now).

smarr:
This is solved with #215.

PR #215, opened 2018-01-18
Support Arbitrary Number of Block Arguments

daumayr:
Expand Blocks so an arbitrary number of arguments can be accepted: `aBlock valueWithArguments: arg` where `arg` is an `Array` containing the arguments.

- change semantics for blocks, so that they throw an error when the number of arguments does not match
- added `#cull*` methods to allow blocks to ignore arguments
- remove the `BlockN` classes, which originated from SOM

This PR solves #94 and #216.


core-lib/Kernel.ns
```diff
@@ -630,18 +630,21 @@ class Kernel vmMirror: vmMirror = Object <: Value (
...
 
   protected class Block3 = Block ()
   (
     (* Evaluating *)
-    public value                  = ( self value: nil with: nil )
-    public value: arg             = ( self value: arg with: nil )
     public value: arg1 with: arg2 = ( ^ vmMirror blockValue: self with: arg1 with: arg2 )
   )
 
+  protected class BlockArbitrary = Block ()
```
smarr: Do we at this point still need the other block classes?
Couldn't we just do away with Block1-3?

smarr: @richard-roberts any opinions/requirements?
How does Grace handle block activations when the number of arguments doesn't match?

core-lib/Kernel.ns
```diff
@@ -64,7 +64,7 @@ class Kernel vmMirror: vmMirror = Object <: Value (
     public notNil = ( ^ true )
     public ifNil: aBlock = ( ^ self )
     public ifNotNil: aBlock = ( ^ aBlock value )
-    public ifNil: noGoBlock ifNotNil: goBlock = ( ^ goBlock value: self )
```
smarr, line 67: This was actually correct. But `ifNotNil:` seems wrong.
The reason is that we want to capture the value and have it in the block argument:

```Smalltalk
obj ifNotNil: [:it | it isGuaranteedToExist ].
(* or also *)
obj ifNil: [ #foo ] ifNotNil: [:it <SomeThing> | it baz ].
```

This is used for instance in MiniTest.

smarr: More digging gives both uses, with, and without `:it`.
Think we need `cull:` then to give a missing value as `nil`.

smarr: Something to consider: http://forum.world.st/cull-protocol-td1560122.html

core-lib/Kernel.ns
ghost: We just throw an error; from the Grace spec: "Requesting the apply(...) method evaluates the block; it is an error to provide the wrong number of arguments."

core-lib/Kernel.ns
daumayr: What is your preference for implementing this: Kernel, Primitive, or both?

smarr: First in Kernel, and if it turns out to be a performance concern, we can optimize it?

daumayr: Sure, we need a numArgs primitive for that anyway, do we want to expose that for other use-cases or just culling?

smarr: Good question. We do not expose it in Symbol either, and I'd consider it reflective information.
So strictly speaking it should be only exposed via mirrors.
Think the best would be to simply use a `vmMirror` primitive directly in the `cull*` implementations.

smarr:
I did a few more cleanups, updated the PR description, and added `#cull*` support.

PR #220, opened 2018-02-11
Improve Files module

smarr:
The goal of this PR is to improve test coverage, correctness, style, API, and functionality of the Files.ns module.

### Main Changes

- changed file mode to be one of `#read`, `#write`, `#readWrite`, add documentation
- add method variants that do not take a failure block, but throw exceptions as common defaults
- fix broken IOException
- simplify `SFileDescriptor`
- make FileTests.ns independent of current directory
- introduce primitive to get file name from class definition
- add more tests
- do not use abbreviated/lower-case class names

### Incidental Changes

- added `ClassDefinitionMirror>>#filePath` to  be able to access the file path of a module
- make `som` and JaCoCo support robust to different current working directory
- fixed issue with Dynamic Metrics tests and coverage tracking


smarr:
@daumayr, I did more work on this. Would be good if you could review, especially to know whether you foresee any issues with what you need.

ghost:
@daumayr @smarr Hoping to use this for Grace, so chop chop :)

smarr:
@richard-roberts where's the review then?

core-lib/Files.ns
```diff
@@ -516,11 +544,25 @@ DAMAGE.>> *)
       ^ self pattern
     )
 
-    public open: mode <Symbol> ifFail: err <[:IOException | X def]>  ^ <FileDescriptor | X> = (
+    (* Open file with given access mode.
+       Access mode can be #read, #write, or #readWrite. *)
+    public open: mode <Symbol> ifFail: errBlock <[:Symbol | X def]> ^ <FileDescriptor | X> = (
       (* Modes: #read, #write, #readWrite *)
```
daumayr: i guess thats no longer needed?

daumayr (approved):
Looks good to me, so much cleaner code when using files when you don't need all those fail blocks.

core-lib/Files.ns
daumayr: * the comment line

smarr: Why? Think it is good practice to provide comments to all methods.
Ideally, VS code is going to show them at some point. (Think we only need to track them during parsing, if we don't do it yet, and then show them on the corresponding request. should be pretty simple)

daumayr: i mean the old one, the content is in the new one, so redundant

smarr: aaah, now I see it, yes :) thanks

smarr:
Going to merge this once CI on GitLab/Brussels finished successfully

PR #222, opened 2018-02-12
Kompos docs

ctrlpz:
Documentation notes about how running Kompos debugger:
- add commands to run the web debugger
- add commands to run the tests

### DIFF
--- a/docs/dev-setup.md
+++ b/docs/dev-setup.md
@@ -89,3 +89,50 @@ A brief list of steps:
 3. Import Eclipse projects
 
 4. Run `ant` from the command line or Eclipse
+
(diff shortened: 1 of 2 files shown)


smarr:
I changed the base for the PR, it was `master` while we need to work against `dev`. Otherwise, there are hundreds of changes in addition.
Will have detailed look later. 

Thanks

smarr:
Hm, wasn't the original thing you wanted to add about how to build Kompos?
These usage instructions are helpful as well, but perhaps in the parts before it, it is useful to point out that the `ant` command will build more than SOMns, i.e., it includes Kompos, which is not included in `ant compile`.

PR #224, opened 2018-02-15
Colorizing terminal output

ghost:
These changes enabled messages printed to standard out to be colorized when they are either a warning or an error.

I've moved all print methods (that were previously in `VM.java`) into a  new `Output.java` module. The module contains print, warning, and error methods that are called whenever SOMns prints something to standard output or standard error. Each method uses the static `AnsiColor8` class to prefix their messages with the corresponding ANSI color code: red for errors, yellow for warnings, and no coloring for regular messages. 

To provide access to the printing methods from Newspeak, I've added a pair of new System primitives: `PrintWarningPrim` and `PrintErrorPim`. These can be invoked, via the System module.
 
Finally, the ANSI coloring is turned off by default (since it's not supported by Eclipse's Console widget). The argument `--ansi-coloring`, or the shorter version `-ac`, can be added when invoking SOMns from the command line to turn the ANSI coloring on.


core-lib/System.ns
```diff
@@ -25,7 +25,8 @@ class System usingVmMirror: vmMirror = Value (
   (* Exiting *)
   public exit: error  = ( vmMirror exit: error )
   public exit         = ( self exit: 0 )
-  public error: msg   = ( self printNewline: 'ERROR: ' + msg. self exit: 1 )
+  public warning: msg   = ( vmMirror printWarning: msg. )
```
smarr: one space too much for alignment before `=`?

src/som/VM.java
```diff
@@ -315,17 +315,43 @@ public void errorExit(final String message) {
     requestExit(1);
   }
 
+  private static class AnsiColor8 {
+    private static String black   = "\u001b[30m";
```
smarr: Is there a specific reason why these are not final?
and, if they should be final, could we capitalize the names, e.g., `BLACK` to indicate them as constants?

smarr: It's also getting unwieldily here.
Could you move all printing into something like an `Output` or `Terminal` class?
The `som` package is probably the right place for it.

smarr (changes requested):
Thanks, I like it pretty :)

Could you please have a look at the comments?
I am also wondering whether this needs to have a flag to disable it, some java property.
When we might run in an environment without ansi term support, this might cause issues.
I checked whether that can be automatically detected, but that seems to be rather problematic.

ghost:
Hi Stefan,

I've updated the changes to include a new argument, added to the `./som` Python script and later VMSettings (I'm happy to change to a Java property if you like, but I thought this was a reasonable solution). When this flag is included, the ANSI coloring is disabled. I didn't know how to name the flag, so currently I've used `-O` for the short name and `--no-ansi` for the long name. 

I've also changed the fields of the `AnsiColor8` class to be public and finalized and refactored the print, warning, and error methods so that other colors can be used.

som
```diff
@@ -105,7 +105,8 @@ parser.add_argument('-T', '--no-trace', help='do not print truffle compilation i
 parser.add_argument('--no-graph-pe', help='disable Graph PE',
                     dest='graph_pe', action='store_false', default=True)
 
-
+parser.add_argument('-O', '--no-ansi', help='Turns off ANSI coloring for output from interperter',
+                    dest='no_ansi_coloring', action='store_true', default=False)
```
smarr: Please avoid negation where possible.
So, ` dest='ansi_coloring', action='store_false', default=True`. This is much more explicit and easier to read.

som
```diff
@@ -224,6 +228,8 @@ if not args.interpreter and args.igv_to_file:
     flags += ['-Dgraal.PrintIdealGraphFile=true']
 if args.igv_methods:
     flags += ['-Dsom.igvDumpAfterParsing=true']
+if args.no_ansi_coloring:
```
smarr: Why are there two flags added?
`-Dsom.noAnsiColoring=true` and `-Dsom.useANSIColor=false`?

src/som/Output.java
```diff
@@ -0,0 +1,82 @@
+package som;
+
+import com.oracle.truffle.api.CompilerDirectives.TruffleBoundary;
+
+import som.vm.VmSettings;
+
+
+public class Output {
```
smarr: Please add a brief comment.

src/som/Output.java
```diff
@@ -0,0 +1,82 @@
...
+  @TruffleBoundary
+  public static void warningPrint(final String msg) {
+    print(msg, AnsiColor8.YELLOW);
+  }
+
+  @TruffleBoundary
+  /**
+   * This method is used to print reports about the number of created artifacts.
+   * For example actors, messages and promises.
+   */
```
smarr: Please first the comment, then the annotation.

src/som/Output.java
```diff
@@ -0,0 +1,82 @@
...
+    public static final String CYAN    = "\u001b[36m";
+    public static final String WHITE   = "\u001b[37m";
+    public static final String RESET   = "\u001b[0m";
+  }
+
+  @TruffleBoundary
+  public static void print(final String msg, final String color) {
+    // Checkstyle: stop
+    if (VmSettings.NO_ANSI_COLOR_IN_OUTPUT) {
+      System.out.print(msg);
```
smarr: This isn't a correct patch.
We used to output to `System.err` when showing errors.

src/som/vm/VmSettings.java
```diff
@@ -29,6 +29,8 @@
 
   public static final boolean IGV_DUMP_AFTER_PARSING;
 
+  public static final boolean NO_ANSI_COLOR_IN_OUTPUT;
```
smarr: No negation, please!

smarr (changes requested):

smarr:
And, I can't see the logs currently because of Travis issues, but I am pretty sure they are failing.
We got tests for a reason :)

ghost:
Hmm, it seems the Eclipse Console has ANSI support now (or perhaps I turned it off somewhere). Maybe the flag should be to turn the coloring on rather than off?

smarr:
> Hmm, it seems the Eclipse Console has ANSI support now (or perhaps I turned it off somewhere). > Maybe the flag should be to turn the coloring on rather than off?

Say what? I don't understand, but this doesn't make sense to me.
Why would you disable it by default if the Eclipse console has support for it?

som
ghost: Haha, missed this during clean-up.

src/som/Output.java
ghost: https://github.com/smarr/SOMns/blob/master/src/som/VM.java#L352

smarr: Yeah, there should be a linter rule for that...

ghost:
> Why would you disable it by default if the Eclipse console has support for it?

Sorry, I had meant to say that Eclipse console does not have support for it.

smarr (approved):

smarr:
Since this is a conceptually pretty simple change, I did condense the history.

PR #225, opened 2018-02-18
Update Truffle to latest version

smarr:
This Truffle update relies on Truffle's new support for `SuspendAnchor` instead of using our own solution based on `SteppingLocation`.
The change seems to be compatible, Kompos tests do not seem to fail.

There are other new features in Truffle, and incompatible changes in Graal/Truffle.
Therefore, anyone trying to use SOMns with this change merged, will need to update GraalBasic to get JIT compilation. (@richard-roberts @daumayr probably relevant for you guys)

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -8,7 +8,8 @@
     <property name="unit.dir"    value="tests/java" />
     <property name="kompos.dir"  value="tools/kompos" />
     <property name="truffle.dir" location="${lib.dir}/truffle/truffle" />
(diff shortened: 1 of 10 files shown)


smarr:
Think I should have fixed all issues now.
Once the tests pass, I'll merge the Truffle update.

This means, you will need to update GraalBasic to the latest version, and if you use Eclipse, you'll need to run `ant ideinit` to get the update Eclipse projects, because they did change the structure of those around.

PR #229, opened 2018-02-23
Throw a proper SOMns Exception when a module isn't found, and fix output coloring

smarr:
This fixes the annoying Java stack trace when a module can't be loaded.
@daumayr since you have been messing with exceptions, you probably want to have a look.

It also includes a fix for `Output`, to always reset colors immediately. @richard-roberts, this seems to be more robust with respect to output generated somehow else, or crashes. We don't want to leave the terminal in a wonky state.


smarr:
@daumayr this should solve the issue you reported on Slack. Will merge once the CI finished.

PR #231, opened 2018-02-24
Use AST Inlining from Black Diamonds

smarr:
With this PR, we are adopting the inlining/splitting support from Black Diamonds: https://github.com/SOM-st/black-diamonds/pull/6

With https://github.com/smarr/SOMns/pull/231/commits/346842b7861f3182a10eee8eeb5b4fa7e219db13, we unfortunately revert https://github.com/smarr/SOMns/pull/177
While there is a performance issue, it is not possible to simply apply the solution in #177. It is semantically not sound, because it results in scopes pointing to the wrong parent. And at the moment, we need a direct parent chain.

The only solution seems to be to revamp scope handling entirely in SOMns. At the moment we mix run-time info and compile-time info (lexical scopes vs. root nodes), which is leading to this issue.
Think, I need to move the run-time info entirely in root nodes or SInvokable/SObject.
How to do that is not entirely clear yet to me. One step in doing this is going to be #233. But there is much more work needed, which I can't do at the moment. So, this means we have a performance regression in the mean time. But it's a somewhat obscure benchmark, so well...

We should also make sure that any issue cleaning this up, removes unnecessary tracking introduced by #177.

The corresponding PR for TruffleSOM is https://github.com/SOM-st/TruffleSOM/pull/13


smarr:
Ok, so, here we go, this should make things work with SOMns.

Unfortunately, this increases warnings with `javac` quite a bit. Can't explain it, the change from before is trivial, except that the classes are now in BD. Might be a `javac` bug.

PR #232, opened 2018-03-07
Simplify setup and use of Graal, and JDK 9 Support

smarr:
Currently, all documentation and the setup is advising people to use [GraalBasic](https://github.com/smarr/GraalBasic), which requires to compile a JVMCI-enabled HotSpot. (JVMCI is the Java Virtual Machine Compiler Interface, which enables the use of Graal)

The main reason for this setup is that it simplifies CI, and allows us to change Graal somewhat independently from the Truffle used by SOMns. This is important to be able to attribute performance changes correctly to the Graal compiler or something else.

Another reason is historical: Graal and Truffle used to be part of separate repositories.
However, now with Graal and Truffle being part of the same repo, we have the situation that SOMns always comes with its own Graal, which we do not actually use.

It seems also too complicated to build HotSpot  for many users.
Thus it would be desirable to avoid this extra hurdle and allow them to use prebuilt binaries.

### Options for how to use Graal

There are a two different options we could take here, and I'd like to hear your opinion: @daumayr @richard-roberts.

#### 1. keep using Graal external to SOMns as default

  **pro:** 
   - not much changes needed, the attached patch does just that. It adds an option to use the embedded Graal with an extra flag
   - can freely update Graal independently of Truffle (more predictable performance on CI, because we know when we change the compiler)

  **con:** it does not make it easier for people to use Graal

#### 2. use the embedded Graal per default
  **pro:** avoids having the Graal source externally, avoids mismatch issues
  **con:** means we either need an option to use external Graal, or bind ourselves to the Graal version of Truffle

~~Even with the support of using the embedded Graal, one hurdle for users is still the need for a JVMCI-enabled Java 8/HotSpot. In the long term, this is going to be fixed by supporting >Java 9.~~

~~In the mean time, it means people need to either compile HotSpot themselves, they need to download it from Oracle Labs, and accept their license, or we need to start building and hosting binaries ourselves. The later is not too complicated, and we could do an automatic download of the right binary depending on the operating system to shield the user somewhat from it.
@richard-roberts was also thinking of improving distribution issues more generally.
But, there remains work to be done.~~

### Java 9 Support

The most recent patch in this PR introduces support to build SOMns on JDK 9.
We do not have full Java 9 support, which would require more work, which currently is not very useful because of a [performance bug in Java](https://bugs.openjdk.java.net/browse/JDK-8189747).
However, with the ability to build on JDK 9, we can advise all users to simply use Java 9, which avoids any separate compilation of HotSpot.

Think this is the best way forward. We use the embedded Graal by default and add support for fallbacks. Note that all testing is still done on Java 8, and only basic functionality is tested on Java 9.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -16,22 +16,22 @@ build_and_test_job:
   stage: build-and-test
   tags: [benchmarks, infinity]
   script:
-    - ant checkstyle
-    - ant eclipseformat-check
-    - timeout 5m ant unit-tests som-tests
+    - ant -Dskip.graal=true checkstyle
+    - ant -Dskip.graal=true eclipseformat-check
+    - timeout 5m ant -Dskip.graal=true unit-tests som-tests
 
 kompos_and_dym_tests:
   stage: full-test
   tags: [benchmarks, infinity]
   script:
-    - ant jacoco-lib compile dynamic-metrics-tests superinstructions-tests
+    - ant -Dskip.graal=true jacoco-lib compile dynamic-metrics-tests superinstructions-tests
     - cd tools/kompos && npm install . && npm -s run verify && npm test
 
 replay_tests:
   stage: full-test
   tags: [benchmarks, infinity]
   script:
-    - timeout 10m ant replay-tests
+    - timeout 10m ant -Dskip.graal=true replay-tests
 
 
 benchmark_savina_job:
@@ -39,7 +39,7 @@ benchmark_savina_job:
   tags: [benchmarks, infinity]
   allow_failure: true
   script:
-    - ant compile
+    - ant -Dskip.graal=true compile
     - export EXP=`if [[ "$CI_BUILD_REF_NAME" = "master" ]]; then echo "SOMns-Savina"; else echo "SOMns-Savina-exp"; fi`; rebench -d --without-nice -c --commit-id="$CI_BUILD_REF" --environment="Infinity Ubuntu" --project=SOMns-Savina --branch=master codespeed.conf $EXP
 
 benchmark_job:
@@ -47,7 +47,7 @@ benchmark_job:
   tags: [benchmarks, infinity]
   allow_failure: true
   script:
-    - ant compile
+    - ant -Dskip.graal=true compile
     - export EXP=`if [[ "$CI_BUILD_REF_NAME" = "master" ]]; then echo "SOMns"; else echo "SOMns-exp"; fi`; rebench -d --without-nice -c --commit-id="$CI_BUILD_REF" --environment="Infinity Ubuntu" --project=SOMns --branch=master codespeed.conf $EXP
 
 benchmark_interp_job:
@@ -55,7 +55,7 @@ benchmark_interp_job:
   tags: [benchmarks, infinity]
   allow_failure: true
   script:
-    - ant compile
+    - ant -Dskip.graal=true compile
     - export EXP=`if [[ "$CI_BUILD_REF_NAME" = "master" ]]; then echo "SOMns-interp"; else echo "SOMns-interp-exp"; fi`; rebench -d --without-nice -c --commit-id="$CI_BUILD_REF" --environment="Infinity Ubuntu" --project=SOMns --branch=master codespeed.conf $EXP
 
 benchmark_nightly_job:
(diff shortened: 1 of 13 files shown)


ghost:
In my opinion, language implementors using SOMns should be able to work with the Truffle DSL without needing to understanding Graal in much detail. As such, my preference would be the latter option of using the embedded Graal. Structuring the repo in this way suggests that Graal is more of a black box than a component the implementor needs to be worried about.

The JVMCI-enabled JDK is still a problem (from my recent experience with @potanin, it seems that this is really the hardest part). If we were to take this path, when could we shift to >Java9?

smarr:
@richard-roberts when ever someone puts in the work to make sure we are Java 9 compatible...

daumayr:
@smarr did i get this correctly: embedded graal basically does what graal-basic does and creates a graal vm in a subdirectory of SOMns?
So the difference between the two options is which jdk location is used.
In the long term option two might be the better choice for people using SOMns, to me it seems changing the flags to switch default case doesn't take that much effort and could be done when demand is high enough.

daumayr:
https://www.oracle.com/corporate/features/migrating-from-java-8-to-java-9.html

daumayr:
https://docs.oracle.com/javase/9/migrate/

smarr:
@daumayr there seems to be confusion about the terminology here.

[GraalBasic](https://github.com/smarr/GraalBasic) is HotSpot (incl. JVMCI) + Graal repo. The embedded Graal is just the Graal repo, so no HotSpot.
The Graal repo contains Truffle, Graal, and Graal SDK.

For running SOMns without JIT compilation, we need Truffle + any JVM.
For running SOMns with JIT compilation, we need Truffle + Graal + a JVMCI-enabled JVM.
I am trying to make it easier here to use a precompiled JVM (with JVMCI support).

I am not trying to create a "GraalVM" in a subdirectory of SOMns.
The GraalVM is Oracle's term for the bundled enterprise version of Graal, HotSpot, SubstrateVM, and their bunch of languages.

daumayr:
With GraalVM i ment the result of compiling GraalBasic, i.e. the contents of ~/.local/graal-basic

smarr:
~/.local/graal-basic is the combination of HotSpot (incl. JVMCI) + Graal, which would not be necessary anymore.

daumayr:
because like with JVMTI you can put JVMCI compilers in the JVM arguments?

smarr:
Yes, indeed. So far, the `build.sh` of GraalBasic simply copied the Graal+GraalSDK jars into the right places, but of we use the embedded one, we can configure them with flags (that's what this PR currently does).

ghost:
So, outside of the CI, are there any advantages of compiling Graal ourselves?

smarr:
@richard-roberts not sure you are referring to what you think you are referring to. In either scenario, Graal is compiled "ourselves" to make sure that it matches our Truffle version. We just don't need to compile HotSpot anymore.

ghost:
Sorry, I meant to say "are there any significant advantages of compiling Hotpot ourselves?"

smarr:
Not much, except that it is easier in the CI, because it doesn't require a download.
Also, for me it was easier...

smarr:
I did spent some time on trying to get things running on Java 9.
Seems problematic. Because of a performance regression for Java's reflection, compilation time goes from 50s to about 7min on my machine, which is absolutely impractical for me.
The issue is documented here, and seemingly only to be solved with Java 11: https://bugs.openjdk.java.net/browse/JDK-8189747

Trying to use the `--release 8` option, i.e., to try compiling as Java 8 seems to be complicated or not even possible, because `sun.misc.Unsafe` isn't easily available anymore.
When trying to work around this with `--add-exports java.base/jdk.internal.misc=ALL-UNNAMED`, we of course get `option --add-exports not allowed with target 1.8`.

Not sure what we can do without having to do major changes, for instance to avoid any unsafe use.

smarr:
Thanks to the help of @eregon, I managed to get things working on Java 9.

With my current set of changes, it should be possible to build SOMns having only Java 9 on the system. We are still build as Java 8 however.

Furthermore, the default behavior will use the embedded Graal compiler, which can be disabled by passing the `-EG` flag to the `som` launcher.

smarr:
Ok, to summarize the latest status (updated the initial issue description, too):

The most recent patch in this PR introduces support to build SOMns on JDK 9.
We do not have full Java 9 support, which would require more work, which currently is not very useful because of a [performance bug in Java](https://bugs.openjdk.java.net/browse/JDK-8189747).
However, with the ability to build on JDK 9, we can advise all users to simply use Java 9, which avoids any separate compilation of HotSpot.

Think this is the best way forward. We use the embedded Graal by default and add support for fallbacks. Note that all testing is still done on Java 8, and only basic functionality is tested on Java 9.

eregon:
@smarr Is Graal bundled with JDK 9 on both Linux and macOS? I seems to recall that was only the case on Linux but it's been a long time. The bundled Graal could also be very old and not compatible with recent Truffle.

smarr:
@eregon don't know. But I am referring to the Graal that's embedded into SOMns. With respect to flags and stuff, it seems to work equally well on Ubuntu and macOS.

eregon:
Ah, that should be fine then. I thought you meant the internal Graal in JDK 9 used for jaotc.

Issue #182, opened 2017-08-07
Simplify/unify scope handling for object literal support


smarr:
The object literal support introduced with #112 avoids changing some of the scope handling infrastructure and instead of modeling lexical scope as a single change, it introduces direct access to scopes of objects.

Specifically, the `MixinBuilder` has not an outer builder and an outer method, it should only have a single outer scope, the rest should be reached through traversal.
Same issue was introduced in the `MixinScope` class, there we also got two scope information now.

This could be hiding bugs, and made it hard to implement the contextual nodes in a way that the parser determines correctly the chain of casts needed to get to the outer context.
Currently, the chain is determined at run time, because I had issues to get it correct during inlining.
The parser support worked, but finding the right data to have the chain correct after splitting was an issue.

smarr:
The first part of this was done with #233, but the second bit could still be investigated.

PR #233, opened 2018-03-12
Handle method/mixin scope uniformly

smarr:
Until now, we kept mixin and method scopes separately (introduced with #112), which does not directly represent the lexical structure and can cause confusion.

With this change, scope is managed as a single chain outwards (solving #182), with methods and mixins on the same chain.

This change applies to LexicalScope classes as well as to the Builder classes.

- add ScopeBuilder superclass for MixinBuilder and MethodBuilder
- correctly determine whether scopes are immutable also for methods
- added Local.isMutable()
- InliningVisitor, handle case that there are no variables

@richard-roberts please test and review this carefully!
Locally, all tests seem to pass, but I expect that your Grace code is more complex, and could hit other issues


ghost:
At first glance, it looks great - something that I can get my head around more easily. I can't get to this now sorry, but will do it on Wednesday.

PR #238, opened 2018-03-16
Fix outer traversal to use lexical chain

smarr:
TODO: implement fix @richard-roberts, this is currently only the test documenting the broken behavior. (because of laziness the PR still contains elements of the release preparation)

This test demonstrates that our traversal of outer objects currently
does follow the bottom of the chain of outer objects, i.e., the outer
object of the most concrete class, instead of following the lexical chain
defined by the lexical scopes.

A solution to this issue probable needs to adapt the outer object read to take the full and correct lexical chain into account.
Roughly something like this:

```java
  @ExplodeLoop
  private Object getEnclosingObject(final SClass lexicalClass) {
    int ctxLevel = contextLevel - 1; // 0 is already covered with specialization
    SObjectWithClass enclosing = lexicalClass.getEnclosingObject();

    while (ctxLevel > 0) {
      ctxLevel--;
      
      SClass cls = enclosing.getSOMClass();
      // TODO: cls = lexicalClass(cls, ctxMixinId[ctx]);
      
      enclosing = cls.getEnclosingObject();
    }
    return enclosingObj.profile(enclosing);
  }
```

### DIFF
--- a/src/som/interpreter/nodes/ResolvingImplicitReceiverSend.java
+++ b/src/som/interpreter/nodes/ResolvingImplicitReceiverSend.java
@@ -27,12 +27,16 @@ public final class ResolvingImplicitReceiverSend extends AbstractMessageSendNode
   /**
    * A helper field used to make sure we specialize this node only once,
    * because it gets removed, and races on the removal are very problematic.
+   *
+   * REM: acccess only under synchronized(this)!
    */
   private PreevaluatedExpression replacedBy;
 
   /**
    * In case this node becomes an outer send, we need to recalculate the
    * receiver also when the specialization was racy.
+   *
+   * REM: acccess only under synchronized(this)!
    */
   private OuterObjectRead newReceiverNodeForOuterSend;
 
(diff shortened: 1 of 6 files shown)


smarr:
@richard-roberts the simplest solution to this issue might be to construct a chain of outer object read nodes. this might even simplify the outer object read node itself. So, each node, steps out exactly once, and then relies on the next one to finish it correctly.

ghost:
@smarr Yes I agree, replacing the logic with a chain of one-step outerReadNodes seems like the cleanest solution. I will prototype it on ASAP and push here.

smarr:
@richard-roberts thanks a lot for fixing this!
The solution looks very good.

To keep the history relevant, I squashed most things into one commit.

I also pushed a few extra changes.
The idea is that we avoid reversing the list of outer ids twice.
The traversal is naturally from the most inner scope to the outer scope. With this in mind, we do not need to reverse the id list twice (once by inserting them into the list at the front, a second time by using a reverse loop).

Going to try to construct a few more tests (I also changed the comment on the existing test, at least one bit was an absolute statement ("can follow any lexical class"), think the test isn't exhaustive enough to justify "any", just yet).

src/som/interpreter/nodes/ResolvingImplicitReceiverSend.java
```diff
@@ -73,38 +73,49 @@ public Object doPreEvaluated(final VirtualFrame frame, final Object[] args) {
...
+
       ExpressionNode[] msgArgNodes = argumentNodes.clone();
-      msgArgNodes[0] = newReceiverNodeForOuterSend;
+      MixinDefinitionId currentMixin = mixinId;
+      MixinDefinitionId enclosingMixin;
+
+      for (int i = result.size() - 2; i > -1; i--) {
+        enclosingMixin = result.get(i);
+
+        newReceiverNodeForOuterSend =
```
smarr: This update is problematic, because it is a transitory update, which is not leading to a correct lookup, but is visible to other threads. Will fix this.

src/som/interpreter/nodes/ResolvingImplicitReceiverSend.java
```diff
@@ -73,38 +73,49 @@ public Object doPreEvaluated(final VirtualFrame frame, final Object[] args) {
...
   }
 
-  private PreevaluatedExpression specialize(final Object[] args) {
+  private PreevaluatedExpression specialize(final VirtualFrame frame, final Object[] args) {
     if (replacedBy != null) {
       // already has been specialized
       if (newReceiverNodeForOuterSend != null) {
         // need to recalculate the real receiver for outer sends
-        args[0] = newReceiverNodeForOuterSend.executeEvaluated(args[0]);
+        args[0] = newReceiverNodeForOuterSend.executeGeneric(frame);
```
smarr: @richard-roberts While I agree that we need this semantics, the change itself is highly problematic, because it relies on complex assumptions to be true.
Generally, we can not reevaluate any nodes arbitrarily, because they might have side effects that are visible on the language level.

Here the key assumption is that the evaluation of the outer chain won't cause side effects.
Specifically, the evaluation of the original `args[0]` is not allowed to have side effects, so that a reevaluation is unproblematic. I have been thinking of ensuring this with assertions, which seems hard. So, I'll put in explicit code that guarantees this.

smarr:
Ok, I'll assume that there is nothing contentious here.
So, will cleanup the commits and then look into merging it.

smarr:
I added a bit more testing, and I think I am happy with merging once tests/benchmarks completed.

ghost:
Thanks for cleaning up the implementation and the comments you added. Looks good to me!

Issue #241, opened 2018-03-23
JDK Bugs with impact on SOMns


smarr:
This issue is meant to track the state of JDK bugs, which have an impact on SOMns.

Currently, I am aware of the following issues:

- [ ] JDK-8189747 [JDK9 javax.lang.model.util.Elements#getTypeElement regressed 1000x in performance](https://bugs.openjdk.java.net/browse/JDK-8189747)
  - Once this is fixed, we should consider the todo in #244.
- [x] JDK-8200166 [Repeating annotations referring to to-be-generated classes don't work](https://bugs.openjdk.java.net/browse/JDK-8200166) (seems to be fixed for JDK 11, not tested)
- [x] JDK-8200163 [javac aborts before first round of annotation processing is started](https://bugs.openjdk.java.net/browse/JDK-8200163) (seems to be fixed for JDK 11, not tested)

Issue #85, opened 2017-01-11
race on object transition


ad1622:
for i in 1 2 3 4 5 6; do ./som -G -at -TF core-lib/Benchmarks/AsyncHarness.som Savina.CigaretteSmokers 50 0  1000:200; done

Run-time Error: EventualMessage failed with Exception.
java.lang.AssertionError: After transitioning the object to a new shape, we expect the layout to be valid. But, this is racy...
	at som.interpreter.objectstorage.InitializerFieldWrite.updateObject(InitializerFieldWrite.java:248)
	at som.interpreter.objectstorage.InitializerFieldWriteNodeGen$UpdateObjectNode_.execute_(InitializerFieldWriteNodeGen.java:1317)
	at som.interpreter.objectstorage.InitializerFieldWriteNodeGen$BaseNode_.acceptAndExecute(InitializerFieldWriteNodeGen.java:123)
	at com.oracle.truffle.api.dsl.internal.SpecializationNode.uninitialized(SpecializationNode.java:407)
	at som.interpreter.objectstorage.InitializerFieldWriteNodeGen$UninitializedNode_.execute_(InitializerFieldWriteNodeGen.java:379)[Thread 1]	A#3		M#1P#3

	at som.interpreter.objectstorage.InitializerFieldWriteNodeGen$BaseNode_.acceptAndExecute(InitializerFieldWriteNodeGen.java:123)
	at com.oracle.truffle.api.dsl.internal.SpecializationNode.removeThis(SpecializationNode.java:247)
	at som.interpreter.objectstorage.InitializerFieldWriteNodeGen$UnwrittenOrGeneralizingValueNode_.execute_(InitializerFieldWriteNodeGen.java:1280)
	at som.interpreter.objectstorage.InitializerFieldWriteNodeGen$BaseNode_.execute0(InitializerFieldWriteNodeGen.java:135)
	at som.interpreter.objectstorage.InitializerFieldWriteNodeGen.executeGeneric(InitializerFieldWriteNodeGen.java:60)
	at som.interpreter.nodes.SequenceNode.executeAllButLast(SequenceNode.java:52)
	at som.interpreter.nodes.SequenceNode.executeGeneric(SequenceNode.java:45)
	at som.interpreter.Invokable.execute(Invokable.java:40)
	at com.oracle.truffle.api.impl.DefaultCallTarget.callDirectOrIndirect(DefaultCallTarget.java:63)
[...]

PR #244, opened 2018-03-24
Reduce race-related stack overflows, and handled invalid layouts in ClassSlotAccessNode

smarr:
If multiple threads mutate objects from the same class, it can happen that the object layout gets invalidated by another thread while we try to access it.

In the worst case, this can lead to stack overflows, because our interpreter tries repeatedly to find a valid object layout while creating dispatch chains, or interacting with the object layout in other nodes, which use recursive fallbacks.

A bad case I saw during debugging was that class loading occurred while initializing a new object layout. This delayed the availability of a valid layout long enough to see stack overflows.

To reduce the chance of this happening, I added a busy loop to wait that our lookup getLayoutForInstancesToUpdateObject() actually returns a valid layout. It might still be invalidate before use again, but hopefully this happens now less often, and prevents stack overflows.

This is likely related to https://github.com/smarr/SOMns/issues/85

@daumayr probably relevant for you, you might still have seen such races in your benchmarks

This change introduces a busy loop, for which we might want to add `Thread.onSpinWait()` once we can use [JDK 9](https://github.com/smarr/SOMns/issues/241).

### DIFF
--- a/src/som/compiler/MixinDefinition.java
+++ b/src/som/compiler/MixinDefinition.java
@@ -653,8 +653,8 @@ public final class MixinDefinition {
       CachedSlotRead read = super.createNode(loc, guard, next, isSet);
       CachedSlotWrite write = loc.getWriteNode(this, guard, next, isSet);
 
-      ClassSlotAccessNode node = new ClassSlotAccessNode(mixinDefinition,
-          read, write);
+      ClassSlotAccessNode node =
+          new ClassSlotAccessNode(mixinDefinition, loc.getSlot(), read, write);
       return node;
     }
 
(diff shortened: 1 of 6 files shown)


daumayr:
Good to hear there is progress on this, surprisingly i haven't had issues with this for a while.

smarr:
The latest changes also solve an issue with not handling invalid layouts in the `ClassSlotAccessNode`.
This could cause data been written to the wrong memory address. While the object layout of the object potentially already has been update.

It is hard to test, but at least I observed an `A` being read where an `R` should be read (both are inner classes in the Parser.ns BasicInterpreterTest). The debugging code is available here: https://gist.github.com/smarr/c7b626dbc8d8c33251d952d6653c2ef3
The following `BasicInterpreterTests` where of good help:

```java
{"Parser", "testOuterInKeyword", 32 * 32 * 32, Long.class, null},
{"Parser", "testOuterWithKeyword", 3 * 4, Long.class, null},
```

By checking whether the read/write operations are valid, we can avoid the issue.
In case they are invalid, we will fall back to the standard accessors on `SObject`, which do the right thing.

PR #250, opened 2018-05-13
Move Actor mailbox loop to Truffle

smarr:
This PR moves the loop that iterates over the messages in a mailbox into Truffle.
This is technically not an optimization in itself, but enables future optimization, because it gives more control over for instance tracing operations.

The PR contains also a number of smaller cleanups.

For reference, I included the code that adds caching for the methods of messages.
However, there are even for simple programs too many messages, so this is not useful, and therefore reverted.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -80,7 +80,7 @@
       </exec>
     </target>
 
-    <target name="truffle-libs" unless="skip.libs" depends="truffle,build-graal">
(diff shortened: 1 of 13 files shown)


smarr:
@daumayr, I started moving some of the things we did for the paper into the main branch.
Would be great if you could have a look, and review the changes.

daumayr:
@smarr, I had a look on the changes,  they look good to me.

smarr:
Thank you. It's merged.

Issue #214, opened 2018-01-17
Optimize Exception signaling


daumayr:
Build a nice exception thrower in Java to avoid helpers in SOMns code.
Introduction of new Exception classes (and throwing them from the Java side) should be clean and easy to do.

smarr:
TODO:

- [x] make sure all `signalNotAValueWith:` errors and similar use only the class of the broken `Value`, but don't let the invalid `Value` object escape. We can't allow references to invalid objects to exist
- [x] make sure we use constant `SSymbol` objects everywhere, the string conversion is just overhead

clementbera:
Attempted to fix in PR:
https://github.com/smarr/SOMns/pull/251

PR #251, opened 2018-05-22
[Issue 214] Exception signaling without helper methods

clementbera:
This PR introduces the infrastructure to do exception signaling without relying on helper methods in the Newspeak code. Thereby, it addresses #214.

It introduces the `ExceptionSignalingNode`, which performs the message sends necessary to signal an exception.
In the interpreter, this allows us to have for instance code like this:

```Java
@Child ExceptionSignalingNode indexOutOfBounds;

// ...

indexOutOfBounds.signal(arr, idx);
```

Work with @daumayr.

This change does not yet fully optimize the exception signaling. For instance, it does not cache the exception class, which could avoid the message send for its lookup. However, this would need to be done very carefully, since the lookup is done via a normal message send, which could have side effects. Thus, it should only be optimized for well known and unproblematic cases.

_[description updated by @smarr to be a summary of the change]_


smarr:
For attribution purposes, I'd add a line like `Co-authored-by: name <name@example.com>` to the commit message, to attribute Dominik. /cc @daumayr 

See
https://help.github.com/articles/creating-a-commit-with-multiple-authors/

There's also a compilation error: https://travis-ci.org/smarr/SOMns/jobs/382258342#L689

core-lib/Kernel.ns
```diff
@@ -954,18 +954,4 @@ class Kernel vmMirror: vmMirror = Object <: Value (
...
-  )
-
-  (* Short cut from VM. Likely to be removed when optimized. *)
-  private signalTOCannotBeValues: aClass = (
-    TransferObjectsCannotBeValues signalWith: aClass
-  )
-
-  private signalArgumentError: message = (
-    ArgumentError signalWith: message.
-  )
```
smarr, line 970: Could we also remove https://github.com/smarr/SOMns/blob/release/core-lib/Files.ns#L664-L672 ?

src/som/interpreter/nodes/ExceptionSignalingNode.java
```diff
@@ -0,0 +1,51 @@
...
+import som.vmobjects.SObject;
+
+
+public class ExceptionSignalingNode extends Node {
+
+  @Child protected ExpressionNode getExceptionClassNode;
+  @Child protected ExpressionNode signalExceptionNode;
+  protected final SObject         module;
+
+  public ExceptionSignalingNode(final String exceptionClassName,
```
smarr: Hm, is that so common that it is worthwhile to have the standard parameter `KernelObj.kernel`?

src/som/interpreter/nodes/ExceptionSignalingNode.java
```diff
@@ -0,0 +1,51 @@
...
+  public Object execute(final Object... args) {
+    SClass exceptionClass =
+        (SClass) ((PreevaluatedExpression) getExceptionClassNode).doPreEvaluated(null,
+            new Object[] {module});
+
+    Object[] allArgs = new Object[args.length + 1];
+    allArgs[0] = exceptionClass;
+    for (int i = 0; i < args.length; i++) {
+      allArgs[i + 1] = args[i];
+    }
```
smarr: Would be nice to extract line 43-47 into a named method

src/som/interpreter/nodes/InstantiationNode.java
```diff
@@ -73,7 +98,8 @@ public SClass instantiateClass(final SObjectWithClass outerObj,
     public static SClass instantiate(final SObjectWithClass outerObj,
         final ClassFactory factory) {
       SClass classObj = new SClass(outerObj, instantiateMetaclassClass(factory, outerObj));
-      return singalExceptionsIfFaultFoundElseReturnClassObject(outerObj, factory, classObj);
+      return signalExceptionsIfFaultFoundElseReturnClassObject(outerObj, factory, classObj,
+          floatingNotAValueThrower, floatingCannotBeValuesThrower);
```
smarr: Hm, can we improve this?

Ideally, the nodes should come from the caller.
The `@Specialization` above could provide instance-specific nodes, and the other callers could pass them in from static fields, if really necessary.

src/som/interpreter/nodes/IsValueCheckNode.java
```diff
@@ -82,8 +81,12 @@ private void dropCheckNode() {
   private static final class ValueCheckNode extends IsValueCheckNode {
     ValueCheckNode(final ExpressionNode self) {
       super(self);
+      thrower = new ExceptionSignalingNode("NotAValue",
+          "signalWith:", self.sourceSection);
     }
 
+    @Child protected ExceptionSignalingNode thrower;
```
smarr: Fields are ideally defined before the constructors

src/som/interpreter/nodes/IsValueCheckNode.java
```diff
@@ -92,7 +95,9 @@ public Object executeEvaluated(final VirtualFrame frame, final Object receiver)
       if (allFieldsContainValues) {
         return rcvr;
       }
-      return KernelObj.signalExceptionWithClass("signalNotAValueWith:", receiver);
+
+      return thrower.execute(rcvr);
+
```
smarr: unnecessary blank line

src/som/primitives/ActivitySpawn.java
```diff
@@ -111,9 +111,13 @@ public final SpawnPrim initialize(final VM vm) {
       this.forkJoinPool = vm.getForkJoinPool();
       this.processesPool = vm.getProcessPool();
       this.threadPool = vm.getThreadPool();
+      thrower = new ExceptionSignalingNode("NotAValue",
```
smarr: We might want factory methods for these common exceptions.
Having all these magic strings all over the codebase is likely not ideal.

src/som/primitives/ActivitySpawn.java
```diff
@@ -111,9 +111,13 @@ public final SpawnPrim initialize(final VM vm) {
...
+          "signalWith:", sourceSection);
       return this;
     }
 
+    @Child protected ExceptionSignalingNode thrower;
```
smarr: Please before constructor

src/som/primitives/ActivitySpawn.java
```diff
@@ -186,9 +190,13 @@ public final SpawnWithPrim initialize(final VM vm) {
...
```
smarr: Please before constructor

src/som/primitives/BlockPrims.java
```diff
@@ -52,11 +52,27 @@ public static final IndirectCallNode create() {
...
   @ImportStatic(BlockPrims.class)
   @Primitive(primitive = "blockValue:", selector = "value", inParser = false,
       receiverType = {SBlock.class, Boolean.class})
-  public abstract static class ValueNonePrim extends UnaryExpressionNode {
+  public abstract static class ValueNonePrim extends UnaryExpressionNode implements ValuePrim {
+
+    @Child ExceptionSignalingNode thrower;
+
+    @Override
+    public void signalException(final String message) {
```
smarr: Hm, can this be done as a default method to avoid code duplication?

src/som/primitives/ExceptionsPrims.java
```diff
@@ -103,4 +103,5 @@ public final Object doException(final SBlock body, final SBlock ensureHandler) {
       }
     }
   }
+
```
smarr: unnecessary  change

src/som/vmobjects/SFileDescriptor.java
```diff
@@ -171,17 +172,7 @@ public void setBufferSize(final int bufferSize) {
...
-    } catch (IllegalArgumentException e) {
-      signalInvalidAccessMode(mode.getString());
-    }
-  }
-
-  public static void signalInvalidAccessMode(final Object mode) {
-    CompilerAsserts.neverPartOfCompilation();
-    KernelObj.signalException("signalArgumentError:", "File access mode invalid, was: "
-        + Objects.toString(mode) + " " + AccessModes.VALID_MODES);
+    this.accessMode = AccessModes.valueOf(mode.getString());
```
smarr, line 175: I don't get this change. Where did the exception go?

smarr (changes requested):
Some comments and potential improvements.

src/som/interpreter/nodes/ExceptionSignalingNode.java
clementbera: There's a dozen senders, but we can remove it if you want to.

core-lib/Kernel.ns
clementbera: Done

src/som/interpreter/nodes/ExceptionSignalingNode.java
clementbera: Done

src/som/interpreter/nodes/IsValueCheckNode.java
clementbera: Done.

src/som/interpreter/nodes/ExceptionSignalingNode.java
smarr: How about a named factory method?
I'd prefer that over a constructor, because it would be more explicit.

So, perhaps `createForKernel(..)`?

src/som/interpreter/nodes/InstantiationNode.java
clementbera: Done

src/som/interpreter/nodes/ExceptionSignalingNode.java
clementbera: Moved to a factory as requested.

src/som/primitives/ExceptionsPrims.java
clementbera: Don't know how to remove that. CheckStyle added it for us I think.

src/som/vmobjects/SFileDescriptor.java
clementbera: In the sender, so it can have the exceptionSignalNode.

src/som/primitives/ActivitySpawn.java
clementbera: Done.

src/som/primitives/BlockPrims.java
clementbera: We can't.
We are not able to use a common superclass.
If we move the code out, methods such as insert are not visible anymore.

core-lib/Files.ns
```diff
@@ -645,6 +645,7 @@ DAMAGE.>> *)
     )
   ) : (
     public signalWith: aMessage = (
+      vmMirror printStackTrace: nil.
```
smarr: That's debugging code, we want to solve this not specific to an exception, but somewhat differently (there is no implementation of stack trace capturing at the moment)

src/som/primitives/SystemPrims.java
```diff
@@ -65,17 +66,19 @@ public final Object set(final SObjectWithClass system) {
...
       MixinDefinition module = vm.loadModule(path);
       return module.instantiateModuleClass();
     } catch (FileNotFoundException e) {
-      PathPrims.signalFileNotFoundException(e.getMessage(), "Could not find module file.");
+      PathPrims.signalFileNotFoundException(e.getMessage(), "Could not find module file.",
+          thrower);
     } catch (NotAFileException e) {
       PathPrims.signalFileNotFoundException(e.getMessage(),
-          "Path does not seem to be a file.");
+          "Path does not seem to be a file.", thrower);
```
smarr: Hm, that's `FileNotFound`?

src/som/primitives/SystemPrims.java
```diff
@@ -65,17 +66,19 @@ public final Object set(final SObjectWithClass system) {
...
     } catch (IOException e) {
-      PathPrims.signalIOException(e.getMessage());
+      PathPrims.signalIOException(e.getMessage(), thrower);
```
smarr: And here the same thrower node is `IOException`? seems fishy

smarr (changes requested):
@clementbera @daumayr  I also just pushed a bunch of changes. They are mostly polish, stuff I didn't see from simple reviewing. Note, this is work in progress. Will continue later.

smarr:
@clementbera @daumayr think this here is a slightly nicer solution, because it avoids an interface: https://github.com/smarr/SOMns/pull/251/commits/52a0041063fd64a7eb5c19b9838233aa63316cf5

Also, I am going to change all `thrower` to use the exception name. Together with another change, the result reads nicely, for instance: `argumentError.signal(...)`

smarr:
Ok, now I see why you might have decided for a lazy approach to initialize those exception nodes.
They need the module, but the module might not yet be loaded. Hmpf.

smarr:
@daumayr @clementbera please review https://github.com/smarr/SOMns/pull/251/commits/8412ba6952057e8b40c2166a29c4d883315d8ce0 carefully (see comment).
This might not be the smallest possible change. However, I consider it a useful exercise of some useful design choices and Truffle techniques. Delayed initialization can be achieved with [node rewriting](https://github.com/smarr/SOMns/pull/251/commits/8412ba6952057e8b40c2166a29c4d883315d8ce0#diff-ef3dce52357724f353c655feea2702c6R101), which keeps the logic localized.

smarr (approved):

smarr:
Ok, I am now happy with this. I did a bunch of other changes and clean ups.
Before merging, I am going to squash most of this, since the history is pretty uninteresting and I prefer a easily understandable set of changes.

Please review carefully.

daumayr:
Looks good to me.
So replacing nodes is an option for lazy initialisation, I didn't know that.

src/som/interpreter/nodes/InstantiationNode.java
```diff
@@ -36,28 +39,37 @@ public static SClass instantiateMetaclassClass(final ClassFactory factory,
...
   }
 
   public abstract static class ClassInstantiationNode extends InstantiationNode {
 
+    @Child protected ExceptionSignalingNode notAValueThrower;
+    @Child protected ExceptionSignalingNode cannotBeValuesThrower;
+
     public ClassInstantiationNode(final MixinDefinition mixinDefinition) {
       super(mixinDefinition);
+      notAValueThrower = insert(ExceptionSignalingNode.createNotAValueNode(
```
clementbera: This was not renamed (still named thrower)

src/som/interpreter/nodes/InstantiationNode.java
```diff
@@ -88,6 +103,11 @@ public SClass instantiateClassWithNewClassFactory(final SObjectWithClass outerOb
 
     public ObjectLiteralInstantiationNode(final MixinDefinition mixinDefinition) {
       super(mixinDefinition);
+      notAValueThrower = ExceptionSignalingNode.createNotAValueNode(
```
clementbera: This was not renamed (thrower) and it seems the insert() function is missing.

src/som/interpreter/nodes/IsValueCheckNode.java
```diff
@@ -80,10 +79,21 @@ private void dropCheckNode() {
...
     ValueCheckNode(final ExpressionNode self) {
       super(self);
     }
 
+    @Override
+    @SuppressWarnings("unchecked")
+    public ValueCheckNode initialize(final SourceSection sourceSection) {
+      super.initialize(sourceSection);
+      notAValue = ExceptionSignalingNode.createNotAValueNode(sourceSection);
```
clementbera: It seems the insert() is missing.

src/som/primitives/arrays/NewImmutableArrayNode.java
```diff
@@ -44,8 +46,17 @@ public boolean matches(final Object[] args, final ExpressionNode[] argNodes) {
...
-  @Child protected IsValue           isValue = IsValueFactory.create(null);
+  @Child protected BlockDispatchNode      block   = BlockDispatchNodeGen.create();
+  @Child protected IsValue                isValue = IsValueFactory.create(null);
+  @Child protected ExceptionSignalingNode notAValue;
+
+  @Override
+  @SuppressWarnings("unchecked")
+  public NewImmutableArrayNode initialize(final SourceSection sourceSection) {
+    super.initialize(sourceSection);
+    notAValue = ExceptionSignalingNode.createNotAValueNode(sourceSection);
```
clementbera: No insert()

clementbera (commented):
I think there are some issues, sometimes nodes are initialized with insert(), sometimes not. I guess they should all have insert. Theres a single place where renaming was not done. I annotated all of them. I wanted to change, but for some reasons I have way too many conflicts when I pull.

smarr:
> I think there are some issues, sometimes nodes are initialized with insert(), sometimes not. I guess they should all have insert. Theres a single place where renaming was not done. I annotated all of them.

Inserts are needed when the nodes are added to a method after the whole method was construct.
Or after the first activation? I am not entirely sure. At one point, all child nodes are adopted ("inserted") automatically.

So, strictly speaking `insert()` is only needed after that point.
And yes, there might be inconsistencies in the code currently.

> I wanted to change, but for some reasons I have way too many conflicts when I pull.

As mentioned before, I started squashing commits that do not add helpful information to the history and only make it harder to understand why and how things changed (fixups, in-development bugs, other changes that are only side-effects of the development, and not relevant for the end result).

smarr:
I finished the final cleanups, squashed the history where useful, and updated the issue title/description to explain the changes.

Note, there is another change hidden in here: the message send nodes for the exception handling are all allocated lazily now. There was some overhead on the startup performance, which is fixed by this.

Will merge once all tests/benchmarks passed.
I also included https://github.com/smarr/SOMns/pull/251/commits/3406c639eaefbfadb6401115872c0b94f2adf27d to add the performance note to `SClass.isKindOf()` (see #253).

PR #252, opened 2018-05-23
Add support for extension modules

smarr:
The main goal of this PR is to allow us to load primitives implemented in Java dynamically at run time.
This is going to be useful to provide support for HTTP, databases, and other things, and avoids to include them into the SOMns codebase directly.

An extension module is loaded the same way as other modules:

```
module:: system loadModule: 'extension.jar' nextTo: self.
```

The result of such a load is a class, as for other modules.
It can be instantiated with `new`, and then provides the primitives as normal Smalltalk messages:

```
m:: module new
db:: m openDatabase: 'file.db'.
db close.
```
(This is just a possible example. Database support is not included in this PR.)

Documentation is added in [docs/extensions.md](https://github.com/smarr/SOMns/blob/extension-modules/docs/extensions.md).

The PR includes a few other minor cleanups, somewhat related to this work.


core-lib/TestSuite/ExtensionTests.ns
```diff
@@ -0,0 +1,64 @@
...
+   other modules.
+*)
+class ExtensionTests usingPlatform: platform testFramework: minitest = Value (
+| private TestContext  = minitest TestContext.
+  private ObjectMirror = platform mirrors ObjectMirror.
+  private FilePath     = platform files FilePath.
+
+  private system = platform system.
+
+  private jarName    = 'extension/build/test-extension.jar'.
```
ghost, line 15: Will the extensions generally be in the same directory? If so, we could further hide the fact that these are written in Java and instead write the name as just `test-extension`. With this approach, it would be simple to swap out the Java module for a Newspeak module later on (without changing any code that uses the replaced module).

docs/extensions.md
```diff
@@ -0,0 +1,113 @@
...
+}
+```
+
+The `IncPrim` node uses `@Primitive` to define a primitive method named `inc`.
+Its implementation simply increments a counter by one.
+Note that the node is a unary node, which means it takes one argument.
+In Newspeak, the first argument, i.e., the receiver is always present,
+but in this case it is simply ignored by the `inc(.)` method.
+
+Note further the use of `@GenerateNodeFactory`, which is ensures that factory
```
ghost: which <del>is</del> ensures that <b>a</b> factory class is generated <b>and</b> that we will use <b>it</b> to create

docs/extensions.md
```diff
@@ -0,0 +1,113 @@
...
+The final element we need is the properties file.
+It needs to be named `somns.extension` and be placed in the root of the JAR.
+It defines which class provides the factories, and in this case it is:
+
+```
+class=ext.Extension
+```
+
+With the primitive nodes, the `Extension` class, and the `somns.extension` file,
+we have everything we need in the JAR file.
```
ghost, line 108: Is it possible, without much change, to allow a jar to provide multiple extension classes? Not sure this would ever be helpful but might be worth thinking about.

src/som/compiler/MixinBuilder.java
```diff
@@ -255,6 +255,14 @@ public void setMixinResolverSource(final SourceSection mixin) {
...
   }
 
+  public ExpressionNode constructSuperClassResolution(final SSymbol superClass,
+      final SourceSection source) {
+    MethodBuilder def = getClassInstantiationMethodBuilder();
+    ExpressionNode selfRead = def.getSelfRead(source);
+    return SNodeFactory.createMessageSend(superClass,
+        new ExpressionNode[] {selfRead}, false, source, null, language);
+  }
+
```
ghost, line 265: Why has this been added / changed?

src/som/vm/ExtensionLoader.java
```diff
@@ -0,0 +1,134 @@
...
+  @Override
+  protected void registerPrimitive(final Primitive prim,
+      final Specializer<VM, ExpressionNode, SSymbol> specializer) {
+    String name = prim.primitive();
+
+    assert !("".equals(name));
+    SSymbol signature = Symbols.symbolFor(name);
+    assert !primitives.containsKey(signature) : "clash of vmMirrorPrimitive names";
+    primitives.put(signature,
+        constructPrimitive(signature, specializer, lang));
```
ghost: Should we prefix primitives by their module names to avoid potential clashes?

core-lib/TestSuite/ExtensionTests.ns
smarr: Same directory is not necessary.
Why do we want to hide the fact that they are written in Java?
That would be a departure from what we currently do: https://github.com/smarr/SOMns/blob/release/core-lib/Platform.ns#L6
`vmMirror load: 'System.ns' nextTo: self`.

Not sure this is useful, and not sure it is desirable.
I would avoid unnecessary magic, I think.

docs/extensions.md
smarr: I would not want to go there at this point.
I would stick with the `one file == one module` as we have it for Newspeak modules. Minimal differences between the two is desirable, I think.

src/som/compiler/MixinBuilder.java
smarr: It's a refactor, extracting common code.

src/som/vm/ExtensionLoader.java
smarr: There are no clashes possible. They are entirely isolated from each other.

core-lib/TestSuite/ExtensionTests.ns
ghost: > Why do we want to hide the fact that they are written in Java?

It seemed like a good idea when I suggested it, but now that I'm trying to write something down I can't come up with an answer. Probably best we ignore this in that case :)

core-lib/Files.ns
```diff
@@ -606,24 +606,24 @@ DAMAGE.>> *)
 
     public readStream ^ <ExternalReadStream> = (
       ^ExternalReadStream onDescriptor: (
-        self open: #read ifFail: [ :err <Exception> | 
-          self error: err
```
daumayr: In our PR, you mentioned "unnecessary change" for such changes. There are a lot of those in this PR. Is this because the formatter changed or something like that?

docs/extensions.md
```diff
@@ -0,0 +1,113 @@
+# Extensions
+
+SOMns aims to be a general purpose language that supports a wide range of use
+cases. Some applications might needs to interact with the underlying system.
```
daumayr: Some applications might needs => Singular or plural?

src/som/vm/ExtensionLoader.java
daumayr: If we are correct, there can be clash only between primitive signatures within a plugin, not between the primitive signature in 1 plugin with the signature in vmMirror or in another plugin.

So why the error msg "assert !primitives.containsKey(signature) : "clash of vmMirrorPrimitive names";"?

A different message could be better, for example ("clash of primitive named " + name  + " in " + moduleName).

This looks important when writing the module to know what is problematic (typically when copy pasting one primitive, modifying the copied version to new code and forgetting to change the primitive signature).

daumayr (changes requested):
How much access do the Java implementations of extensions have to the SOMns implementation?
For example when tracing external Data, can I just import and use ActorExecutionTrace, or do i have to pass a reference to the extension somehow?

Is there a dependency system planned for the plugins? Assuming the plugin java code can import anything from the SOMns core codebase, is it possible for a plugin to also import classes from another plugin?

smarr:
> How much access do the Java implementations of extensions have to the SOMns implementation?
> For example when tracing external Data, can I just import and use ActorExecutionTrace, or do i have to pass a reference to the extension somehow?

Extensions are currently loaded into their own class loader. Thus, there isn't really much interaction between the interpreter and the extension at this point.
The normal class loader is the parent, so, the extension might see the interpreter state.
I am not entirely sure how the isolation between class loaders works.
So, that probably should be tested.
If we need to make things accessible, this could probably done via the `Extension` interface.
So, we kind of need to know what's needed.


> Is there a dependency system planned for the plugins? Assuming the plugin java code can import anything from the SOMns core codebase, is it possible for a plugin to also import classes from another plugin?

Importing? Well, class loading logic applies.
If a class is on the classpath, you get access to it.
If it doesn't, like it is the case for other extension, you don't. (Also, please do not use a new term. I don't know what a plugin is. I don't think I use "plugin" anywhere.)
I am not sure why we would want the complexity of making extension modules aware of other modules beside by using standard mechanisms (passing in references as needed).
You got a specific use case in mind?

core-lib/Files.ns
smarr: Yeah, I blame this change on you. You should set your editor to trim unnecessary whitespace at the end of lines ;)

smarr:
I did some more work on this.

@daumayr since you requested changes, could you please confirm all issues are solved, and request new changes or approve the PR?

daumayr:
@smarr I think it would be good to define how library dependencies of the extensions should be handled.
For example for a database that needs the Derby jar, should the extensions provide a setup script?
I don't see any other issues.

smarr:
>  I think it would be good to define how library dependencies of the extensions should be handled.

Hm, you mean in terms of other class path dependencies only? Perhaps this could work: https://docs.oracle.com/javase/tutorial/deployment/jar/downman.html

> should the extensions provide a setup script?

What would the setup script do?

daumayr:
1) Yes, this looks like a solution. Do we want a uniform directory structure for jar dependencies of extensions? 

2) The script would for example download derby from the Appache server and put in the right location.

smarr:
> Yes, this looks like a solution. Do we want a uniform directory structure for jar dependencies of extensions?

Probably, to ease maintenance. Is that something that needs to be enforced somehow?

> The script would for example download derby from the Appache server and put in the right location.

Isn't that something the build script needs to do anyway? I am not sure why you need a separate 'script' for that.

daumayr:
> Probably, to ease maintenance. Is that something that needs to be enforced somehow?

Maybe put a best practice/guideline/convention in the doc, enforcement doesn't seem necessary right now.

> Isn't that something the build script needs to do anyway? I am not sure why you need a separate 'script' for that.

Right

smarr:
@daumayr could you please take another look? (your review still indicates that changes are requested)
I added some more documentation, and a test for using the `Class-Path` attribute of a JAR.
If that's all that we need at the moment, I'd merge this.

daumayr (approved):
I don't see any issues

PR #254, opened 2018-06-13
Tracing and Replay Infrastructure

daumayr:
This PR includes the new efficient tracing and replay, which is able to handle non-deterministic inputs.
The previous tracing (Kompos protocol) is still available as Medeor tracing during replay executions.


core-lib/Actors.ns
```diff
@@ -151,6 +152,14 @@ class Actors usingVmMirror: vmMirror usingKernel: kernel = Value (
     )
   )
 
+  private class Timer = Value (
```
smarr: Why is this class necessary? Does it do anything?

Also, why is this a value?
Is it really appropriate to share timers between actors? This sounds like sharing race conditions between actors.

core-lib/Actors.ns
```diff
@@ -151,6 +152,14 @@ class Actors usingVmMirror: vmMirror usingKernel: kernel = Value (
...
+    vmMirror actorSetupTimer: nil.
```
smarr: Why is this done for every instantiation?
The code looks like it would need to run exactly once, no?

Could you just check whether stuff is initialized in `actorDo:after:`?

core-lib/Actors.ns
```diff
@@ -258,9 +267,6 @@ class Actors usingVmMirror: vmMirror usingKernel: kernel = Value (
   )
 
   public after: milliseconds do: aBlock = (
-    | promisePair |
-    promisePair:: createPromisePair.
-    promisePair promise whenResolved: aBlock.
-    vmMirror actorResolveProm: (promisePair resolver) after: milliseconds.
+      timer <-: after: milliseconds do: aBlock.
```
smarr: Why is this an async send? This is going to register a callback, why can't this be synchronous?
Again, why do you need a timer object here?

core-lib/Benchmarks/AsyncHarness.ns
```diff
@@ -120,6 +120,9 @@ class AsyncHarness usingPlatform: platform = Value (
     )
 
     printRun: runTime = (
+      | stats = system traceStatistics. |
+      (name + ': trace size:    ' + (stats at: 1) + 'byte') println.
+      (name + ': external data: ' + (stats at: 2) + 'byte') println.
       (name + ': iterations=1' + ' runtime: ' + runTime + 'us') println
```
smarr, line 126: It would be nice if the trace size/external data info is only printed when a flag was given to the harness.

src/som/VM.java
```diff
@@ -46,8 +46,9 @@
 import som.vmobjects.SClass;
 import som.vmobjects.SObjectWithClass.SObjectWithoutFields;
 import som.vmobjects.SSymbol;
-import tools.concurrency.ActorExecutionTrace;
+import tools.concurrency.MedeorTrace;
```
smarr, line 49: MedeoTrace?

This seems to be the wrong name.
My understanding is that the "Debugger" elements in the interpreter are called Medeor, so, that should perhaps be a tools.medeor package.
However, the trace here, refers probably to the kompos protocol, and as such, should probably name after the protocol and not the subsystem implementing it.

src/som/VM.java
```diff
@@ -286,16 +287,15 @@ public void shutdownAndExit(final int errorCode) {
 
     shutdownPools();
 
-    Actor.reportStats();
```
smarr, line 289: This is not relevant anymore for the Kompos traces?

src/som/VM.java
```diff
@@ -339,8 +339,11 @@ public void initalize(final SomLanguage lang) throws IOException {
     mainActor = Actor.createActor(this);
     vmMirror = objectSystem.initialize();
 
-    if (VmSettings.ACTOR_TRACING) {
-      ActorExecutionTrace.recordMainActor(mainActor, objectSystem);
+    if (VmSettings.ACTOR_TRACING || VmSettings.MEDEOR_TRACING) {
```
smarr, line 342: `MEDEO_TRACING` should probably be `KOMPOS_TRACING`.

Any idea for a name for your stuff?
If no, perhaps simply `RECORD_REPLAY_TRACING`?

smarr (changes requested):
A first partial and superficial look.

PR #256, opened 2018-07-10
Use CompilerDirectives::castExact instead of ValueProfiles

### LINKED ISSUE #255: Array storage Strategies: Try to replace value profile with CompilerDirectives.castExact
Should be possible.
Thanks @fniephaus for the hint.
Essentially replace https://github.com/smarr/SOMns/commit/e2431b2acc3e31c4e512157e110795943f2a36f2

This change should be carefully benchmarked.

fniephaus:
The assumption is that performance should remain the same (benchmarks needed for verification). At the same time, this change would simplify code.

Closes #255

### DIFF
--- a/src/som/interpreter/actors/TransferObject.java
+++ b/src/som/interpreter/actors/TransferObject.java
@@ -14,7 +14,6 @@ import som.interpreter.objectstorage.StorageLocation;
 import som.vm.NotYetImplementedException;
 import som.vm.constants.Nil;
 import som.vmobjects.SAbstractObject;
-import som.vmobjects.SArray;
 import som.vmobjects.SArray.PartiallyEmptyArray;
 import som.vmobjects.SArray.STransferArray;
 import som.vmobjects.SObject;
@@ -100,7 +99,7 @@ public final class TransferObject {
     transferMap.put(arr, newObj);
 
     if (newObj.isObjectType()) {
-      Object[] storage = newObj.getObjectStorage(SArray.ObjectStorageType);
+      Object[] storage = newObj.getObjectStorage();
 
       for (int i = 0; i < storage.length; i++) {
         Object orgObj = storage[i];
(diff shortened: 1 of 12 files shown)


fniephaus:
Turns out the Truffle version that SOMns is using, [does not implement castExact yet](https://github.com/smarr/truffle/blob/3ed852dd7266e1423e848b64673f906da2e70ed7/truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/CompilerDirectives.java).

smarr:
:( damn, yeah, I need to update it, but that's going to take time, which I currently don't have.

smarr (approved):
Thanks for the PR!

The changes look good to me.
But yeah, I won't be able to merge this before updating Truffle.

And, updating Truffle means I need to update my Truffle changes, adapt the changes for the instrumentation, etc. Last time I tried doing that, I did not manage to get everything working again. So, will take more time.

fniephaus:
No worries, feel free to merge this when done updating Truffle :)
BTW: I've adjusted GraalSqueak and haven't noticed any significant change in performance.

smarr:
@fniephaus Hm, I guess we could do an in-between change, but not sure this is super useful at the moment.

smarr:
And for the record, the `CompilerDirectives.castExact(storage, int.class);` is not correct, because we have an object. Therefore, it should be `CompilerDirectives.castExact(storage, Integer.class);` (as seen in the gitter chat)

fniephaus:
Why shouldn't this be useful at the moment? Performance should remain the same, code is simpler? castExact will only help to remove the remaining ValueProfile. Also, we don't need to it for the plain Integer case, A cast to an unboxed int should do the trick, right?

smarr:
I fixed a couple of compilation issues, and applied the exactCast. Tests are passing, but still waiting for the benchmark results.

smarr:
Results are very noisy. Ups and downs, on peak performance as well as interpreter speed.
Think the measurements aren't very reliable.
Though, they don't indicate a performance problem.
So, I'll merge this since it simplifies the code.

PR #257, opened 2018-07-12
Preparations for new Tracing/Replay implementation

daumayr:
This PR contains preparation for the new tracing & replay.
The custom ByteBuffer is only temporary and will be removed in the next PR.
Please note that replay is not functional as of this PR.

Due to the split into two tracing variants, a lot of code moved to different files.


src/som/primitives/SystemPrims.java
```diff
@@ -45,13 +45,17 @@
...
 import som.vm.constants.Classes;
 import som.vm.constants.Nil;
 import som.vmobjects.SArray;
 import som.vmobjects.SArray.SImmutableArray;
 import som.vmobjects.SObjectWithClass;
 import som.vmobjects.SSymbol;
 import tools.SourceCoordinate;
+import tools.concurrency.ActorExecutionTrace;
+import tools.concurrency.TraceParser;
+import tools.concurrency.TracingBackend;
```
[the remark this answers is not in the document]
smarr: @daumayr Please fix these issues.

smarr:
After fixing the infrastructure issues (which seemed trivial and I bet it would have been easy to figure out), the Kompos tests are failing.

How do you want to play this? Are you going to look into the issues?

src/som/primitives/StringPrims.java
```diff
@@ -150,6 +150,7 @@ protected boolean isTaggedWithIgnoringEagerness(final Class<?> tag) {
       }
     }
 
+    @TruffleBoundary
```
smarr: Hmm, why did you add a boundary here?
The code looks fine. It should compile, no?

daumayr: It's probably a relict of changes that are not in this PR.
If I have an issue with this at some point , I'll solve it differently.

smarr: I just wrote a microbenchmark to see whether and how it compiles.
And I looked at the IGV graphs. It compiles surprisingly good, despite the three times bounds checking (SOMns, Java string, char array).

So, I would expect it to be safe to remove the boundary here. (will do)

src/som/interpreter/actors/Actor.java
```diff
@@ -298,16 +285,16 @@ private void execute(final EventualMessage msg,
         TracingActor.handleBreakpointsAndStepping(msg, dbg, actor);
       }
 
-      try {
-        if (VmSettings.ACTOR_TRACING) {
-          ActorExecutionTrace.scopeStart(DynamicScopeType.TURN, msg.getMessageId(),
-              msg.getTargetSourceSection());
-        }
-        msg.execute();
-      } finally {
```
smarr: Is the removal of this finally intentional? Seems to be the safer to use a finally here.

smarr:
@daumayr ok, it is merged. Looking forward to the other parts.

PR #260, opened 2018-07-19
Ensure charAt: and File read/write are compatible with PE

smarr:
This PR adds microbenchmarks for `String>>#charAt:` and file read/write.

It also fixes basic compilation issues in file reading and writing.

@daumayr please review.


daumayr (approved):
Thanks for fixing this, changes look good

smarr:
There is something broken in `brew`. Not sure what to do about it beside ignoring the failure for the moment. My attempts to ensure the most recent brew is installed do not seem to be sufficient.

So, will ignore the failing Travis run for now.

PR #261, opened 2018-07-23
Optimized Tracing and Restoring Replay Support

daumayr:
This PR contains the second half of the changes to the tracing implementation.
The changes include a move away from ByteBuffer, instead byte arrays and unsafe are used.
Furthermore, Truffle Nodes are used for tracing to achieve lower overhead.


src/som/interpreter/actors/Actor.java
```diff
@@ -302,9 +302,6 @@ private void execute(final EventualMessage msg,
         if (VmSettings.KOMPOS_TRACING) {
           KomposTrace.scopeEnd(DynamicScopeType.TURN);
         }
-        if (VmSettings.ACTOR_TRACING) {
-          ActorExecutionTrace.recordMessage(msg);
-        }
```
smarr, line 300: Is this removal correct?

daumayr: yes, messages are now traced in ReceivedRootNode.

smarr (approved):
Thank you.
Looks good to me.

Will merge once the benchmarks completed.

Issue #262, opened 2018-07-23
Trace TimerPrim


daumayr:
Trace TimerPrim to make it deterministic in replay.
Remember to enable the non-deterministic validation tests

smarr:
@daumayr this is solved, and completed with #264, right?

daumayr:
yes, the tests have been enabled

PR #264, opened 2018-07-24
Redesign Timer Prim

daumayr:
This PR adopts the tracing of non-deterministic inputs for the TimerPrims 
to make timed behaviour deterministic in replay executions.
The mechanism of TimerPrims was changed from resolving a Promise to sending a (external) Message that targets a Block.
The Validation replay tests are re-enabled as they now pass.

This solves Issue #262


src/som/primitives/TimerPrims.java
```diff
@@ -0,0 +1,171 @@
...
+  public abstract static class SetupTimerPrim extends UnarySystemOperation {
+    @Specialization
+    protected final Object setup(final Object o) {
+      timer = new Timer();
+      actorPool = vm.getActorPool();
+      language = vm.getLanguage();
+      timerActor = vm.getMainActor();
+      SInvokable s =
+          (SInvokable) Classes.blockClass.lookupMessage(SELECTOR, AccessModifier.PUBLIC);
+      rct = createOnReceiveCallTarget(SELECTOR,
```
smarr: Any reason not to use the `ValueNonePrim` directly here?

Think, I'd just use that. Will perhaps also reduce code complexity here.

smarr: Hm, I see, `AbstractMessageSendNode` is expected by `ReceivedMessage`.

smarr (approved):
I simplified the change a little, and change the naming of variables to make the code more explicit.

PR #266, opened 2018-08-30
Update Truffle to 1.0.0-rc5

smarr:
This is an update to Truffle >1.0.0-rc5.

@daumayr @clementbera I'd appreciate if you could take a look.
Because of the huge number of changes, I don't really expect a complete review, but would be good if you'd be aware in which areas changes happened, and whether it may effect your work.

It took me quite a while to update SOMns, because some of our patches to Truffle needed
major revisions or reimplementation to account for some new features in Truffle.

Good news is that there are now fewer patches needed:

 - we can now use `Node.notifyInserted(.)` to announce new or changed tags 
 - we can now use `InstrumentableNode.hasTag(.)` to query for tags (only necessary in tools)
 - the new polyglot `Context` doesn't do strict thread checks anymore, we don't need to hack around it anymore.

Some patches required a reimplementation, or were newly required:

- we use tags for advance breakpoints, for instance on thread creation, on joining a fork join task, or sending an actor message: https://github.com/smarr/truffle/commit/d448d9c2af94c7f9d771139123d96209a10361e1
 - we use a stepping strategy to step until the next node with a given tag.
This is very useful when I can't statically determine the target of a specific
stepping operation, but I know that it has to be of a specific kind.
We use this for instance to trigger a breakpoint when an async. message/callback
is starting to execute, or just before a method returns that is going to be
used to resolve a promise. Highly useful to step through async. code.
https://github.com/smarr/truffle/commit/7689d6f383621959d357dfe492bcee85ae3eb466
 - much of this depends on being able to set breakpoints precisely.
So, we also needed support for that. Currently, Truffle tries to find a
best match, perhaps based on column and line, which isn't precise enough for
my use case:
https://github.com/smarr/truffle/commit/3c68e3dcffb3beab3e537ce5ac8e61a78c6b449d

# Other Update-related Changes

## Adopting `@GenerateWrapper`

Truffle switched to a new approach for wrapper nodes for instrumentation.
They are now created by `@GenerateWrapper` and whether something is instrumentable is determined via an interface.
Beside rewriting our custom Truffle patches, the change that required most work was likely
going from @Instrumentable to @GenerateWrapper.
Details: https://github.com/smarr/SOMns/pull/266/commits/7c21cda8d69ac479dfa0d838061a049fb02311ce

To make this work nicely, I had to change a couple of things around.
One problem with the notion of wrappers is that it is rather complicated to
ensure that they always have the right type. 
The type is constrained on the one side by the field containing the child node,
and on the other side, by the node that is actually wrapped.
So, there is a tension between having few very general wrappers, and the
specifics one might want to use in some node.
It requires quite a bit of care to have a more specific field type for the
child node and make everything work out correctly.

## Truffle Interop

The language interop support is something I haven't invested much time in yet.
So, the fact that Truffle's old interop testing infrastructure is gone is somewhat 
unfortunate, because this means interop support is now untested.
There is only a WIP commit: https://github.com/smarr/SOMns/pull/266/commits/8c419709c5243282cdd81abd575afd59342bbd15

## GraalVM Launcher

To account for the new Context and GraalVM changes, I adopted a GraalSDK style launcher, but it's very basic: https://github.com/smarr/SOMns/pull/266/commits/d338791c514fb0cc89765e7198efd95716e9bc74

# Work In Progress/Needs More Work

- CPU Sampler doesn't have any output, while the old profiler had a histogram:  8a589d82fa3b0d4284680912fbe9678c99fec640
- Truffle Interop TCK needs work to be supported: 8c419709c5243282cdd81abd575afd59342bbd15


smarr:
With this being merged, please note that submodules will need to be updated.

PR #268, opened 2018-09-03
Fix replay test flakiness

daumayr:
Exit code was changed to 1 when there were some unprocessed but expected messages at termination.
Exit code no longer touched by message check, message reports only printed when exiting with codes other than 0.


smarr (approved):
Looks good to me.

The code in `replayCanProcess(.)` looks a little ugly. Perhaps there is a simple method hiding somewhere to make it look nicer. I won't bother in this PR though.

Thanks.

PR #269, opened 2018-09-11
Fixes potential NullPointerException in ExpressionNode::toString

spilot:
The NPE is thrown when SourceSection == null

### DIFF
--- a/src/som/interpreter/nodes/ExpressionNode.java
+++ b/src/som/interpreter/nodes/ExpressionNode.java
@@ -160,6 +160,6 @@ public abstract class ExpressionNode extends SOMNode {
 
   @Override
   public String toString() {
-    return getClass().getSimpleName() + "(" + sourceSection.toString() + ")";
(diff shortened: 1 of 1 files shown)


smarr (approved):
Good catch. Thank you!

smarr:
Just a bit on process:

I changed the base for the PR to the `dev` branch. Usually, we keep `release` unchanged, even for bug fixes. Our project isn't large enough to be able to maintain more than one branch unfortunately.

And, then a tip for using branches/github: if you would push the change to a separate branch, it is easier to iterate on more complex changes, and then simply push to that branch. After it is merged, it can also be simply deleted again. Another benefit is that your own main development branch doesn't depend on the progress of the PR.

As for this change, it looks all good. Tests simply failed, because the release branch currently fails building. Think that's an issue with missing libraries, things that moved on the internet. Checkstyle for instance.

spilot:
Got it, thank you! :)

PR #271, opened 2018-09-26
Basic Object Serialisation

daumayr:
This PR is the first step towards introducing actor snapshots to SOMns,
it adds basic (de)serializing operations and infrastructure that will be needed 
in the creation of snapshots. 
Currently serialisation is limited to acyclic object graphs without duplicates (i.e. a tree).

Each SClass now has a serialization node, which can be used to (de)serialize instances of the represented class.  
Tests are based on cloning objects by serialisation & deserialisation.

Possibly introduces a performance regression for the startup time of the List benchmark on Graal (#273)


src/tools/snapshot/nodes/ArraySerializationNode.java
```diff
@@ -0,0 +1,144 @@
...
+  @Override
+  public void serialize(final Object o, final SnapshotBuffer sb) {
+    assert o instanceof SArray;
+    SArray sa = (SArray) o;
+    serializeStorage(sa, sb);
+  }
+
+  protected void serializeStorage(final SArray sa, final SnapshotBuffer sb) {
+    int requiredSpace;
+    if (sa.isBooleanType()) {
```
smarr: Since this looks like a node, shouldn't it simply specialize on the array type with a `@Specialization` and guard?

src/tools/snapshot/nodes/BlockSerializationNode.java
```diff
@@ -0,0 +1,168 @@
...
+      int j = 0;
+
+      sb.putByteAt(base, (byte) fd.getSlots().size());
+      base++;
+      for (FrameSlot slot : fd.getSlots()) {
+        // assume this is ordered by index
+        assert slot.getIndex() == j;
+
+        Object value = mf.getValue(slot);
+        switch (fd.getFrameSlotKind(slot)) {
```
smarr, line 75: Hm, perhaps an opportunity for nodes to optimize, too.
Am also wondering whether the normal local variable read nodes can be reused, not sure it's worth it though.

Also note that we do not support all kinds, we don't have byte, float, and int values for instance.

src/tools/snapshot/nodes/ObjectSerializationNode.java
```diff
@@ -0,0 +1,236 @@
...
+      layout = clazz.getLayoutForInstancesUnsafe();
+      fieldReads = insert(reads);
+      fieldCnt = fieldReads.length;
+    }
+
+    @ExplodeLoop
+    @Override
+    public void serialize(final Object o, final SnapshotBuffer sb) {
+      assert o instanceof SObject;
+      ((SObject) o).updateLayoutToMatchClass();
```
smarr: This seems to be something that should be done conditionally, as part of specialization.

Also, this should really be done in a safepoint, see https://github.com/smarr/SOMns/blob/dev/src/som/interpreter/nodes/dispatch/UninitializedDispatchNode.java#L160

src/tools/snapshot/nodes/ObjectSerializationNode.java
```diff
@@ -0,0 +1,236 @@
...
+      for (int i = 0; i < fieldCnt; i++) {
+        Object value = fieldReads[i].read((SObject) o);
+        // TODO type profiles could be an optimization (separate profile for each slot)
+
+        // TODO optimize, maybe it is better to add an integer to the objects (indicating their
+        // offset) rather than using a map.
+
+        if (!sb.containsObject(value)) {
+          // Referenced Object not yet in snapshot
+          SClass c = Types.getClassOf(value);
```
smarr: For performance, we likely want to start caching these things.

build.xml
```diff
@@ -444,7 +453,7 @@
 
     <target name="coverage" depends="send-java-coverage,send-somns-coverage" />
 
-    <target name="tests" depends="core-tests,replay-tests,coverage">
+    <target name="tests" depends="core-tests,serialization-tests,replay-tests,coverage">
```
smarr, line 458: This target is not tested on Travis or GitLab CI. Depending on how long the tests takes, I'd add them to one of the existing test jobs.
Possibly to one of the cases here: https://github.com/smarr/SOMns/blob/dev/.travis.yml#L67
`ant serialization-tests` or the call to `./som ...` directly.

core-lib/Actors.ns
```diff
@@ -159,6 +159,14 @@ class Actors usingVmMirror: vmMirror usingKernel: kernel = Value (
     ^ vmMirror actors: nil createFromValue: aValue
   )
 
+  public snapshot: obj = (
+    vmMirror snapshot: obj.
+  )
+
+  public sclone: obj = (
```
smarr: No abbreviations please.

core-lib/TestSuite/Serialization.ns
```diff
@@ -0,0 +1,193 @@
...
+  private Actors = platform actors.
+  private String      = platform kernel String.
+  private Array = platform kernel Array.
+  private TransferArray = platform kernel TransferArray.
+  private ValueArray = platform kernel ValueArray.
+  private Vector = platform kernel Vector.
+|)(
+
+  public class ValueClass new: a and: b = Value (
+      |
```
smarr: Inconsistent whitespace format for fields.

core-lib/TestSuite/Serialization.ns
```diff
@@ -0,0 +1,193 @@
...
+    )(
+      public verify: clone =(
+        con assert: f1 equals: clone f1.
+        con assert: f2 equals: clone f2.
+      )
+    )
+
+  public class SerializationTest = TestContext (
+    |
```
smarr: Inconsistent whitespace format for fields.

core-lib/TestSuite/Serialization.ns
```diff
@@ -0,0 +1,193 @@
...
+      cloneAndVerify: original.
+    )
+
+    public testClassReference = (
+      | original clone |
+      original:: SerializationTest.
+      clone:: Actors sclone: original.
+      original println.
+      clone println.
+      assert: original == original.
```
smarr: `#assert:is:` would be better here, I think.

Also, shouldn't this be comparing original to clone?

core-lib/TestSuite/Serialization.ns
```diff
@@ -0,0 +1,193 @@
...
+      original f5 at: 3 put: 1.
+      original f5: EmptyClass new.
+      cloneAndVerify: original.
+    )
+
+    public testClassReference = (
+      | original clone |
+      original:: SerializationTest.
+      clone:: Actors sclone: original.
+      original println.
```
smarr: there shouldn't be prints in tests.

src/som/compiler/MixinDefinition.java
```diff
@@ -850,4 +850,14 @@ public ForeignAccess getForeignAccess() {
   public static boolean isInstance(final TruffleObject obj) {
     return obj instanceof MixinDefinition;
   }
+
+  public String getIdentifier() {
```
smarr: Could you unify this with `ClassFactory.getFullyQualifiedName`?
At the moment, this looks like duplicated code. I'd think we won't need `ClassFactory.getFullyQualifiedName` any longer with this?

src/som/compiler/MixinDefinition.java
```diff
@@ -850,4 +850,14 @@ public ForeignAccess getForeignAccess() {
...
+    MixinDefinition outer = getOuterMixinDefinition();
+    if (outer != null) {
+      return outer.getIdentifier() + "." + this.name.getString();
+    } else if (this.isModule && this.sourceSection != null) {
+      return this.sourceSection.getSource().getPath() + ":" + this.name.getString();
```
smarr: Is this an absolute path? Might be problematic.

src/som/primitives/SystemPrims.java
```diff
@@ -315,6 +320,59 @@ public final long doSObject(final Object receiver) {
     }
   }
 
+  /**
+   * This primitive serves testing purposes for the snapshot serialization by allowing to
+   * serialize
+   * objects on demand.
+   *
+   * @author dominikaumayr
```
smarr: We don't have the `@author` tag anywhere else, and also don't have a free line after the comment.

src/som/primitives/SystemPrims.java
```diff
@@ -315,6 +320,59 @@ public final long doSObject(final Object receiver) {
...
+      if (VmSettings.SNAPSHOTS_ENABLED) {
+        SnapshotBuffer sb = new SnapshotBuffer();
+
+        if (!sb.containsObject(receiver)) {
+          SClass sc = Types.getClassOf(receiver);
+          sc.getSerializer().serialize(receiver, sb);
+        }
+        return 0;
+      }
+      return 0;
```
smarr: Why do you return `0`?

Wouldn't the `receiver` be more usual?

Also, the multiple return statements seem redundant.

src/som/primitives/SystemPrims.java
```diff
@@ -315,6 +320,59 @@ public final long doSObject(final Object receiver) {
...
+        SnapshotBuffer sb = new SnapshotBuffer();
+
+        if (!sb.containsObject(receiver)) {
+          SClass sc = Types.getClassOf(receiver);
+          sc.getSerializer().serialize(receiver, sb);
+          ByteBuffer bb = sb.getBuffer();
+
+          short cId = bb.getShort();
+
+          SClass clazz = SnapshotBackend.lookupClass(cId);
```
smarr: is `clazz != sc`? Seems strange.

I'd expect an assert here that they are the same, even if you feel the need to do the lookup.

Hm, or they change identity?
But at least the underlying mixin should be the same, right?

src/som/vm/constants/Classes.java
```diff
@@ -34,31 +49,59 @@
...
     // Allocate the Metaclass classes
     metaclassClass = ObjectSystem.newMetaclassClass(KernelObj.kernel);
+    metaclassClass.setSerializer(ClassSerializationNode.create());
     classClass = new SClass(KernelObj.kernel);
+    classClass.setSerializer(ClassSerializationNode.create());
     SClass classClassClass = new SClass(KernelObj.kernel);
     ObjectSystem.initializeClassAndItsClass("Class", classClass, classClassClass);
 
     // Allocate the rest of the system classes
+
```
smarr: why an extra blank line?

src/som/vm/constants/Classes.java
```diff
@@ -34,31 +49,59 @@
...
     topClass = ObjectSystem.newEmptyClassWithItsClass("Top");
+    topClass.setSerializer(new SObjectWithoutFieldsSerializationNode(topClass));
```
smarr: Hmm, I am not super happy about making this whole method more complex/longer.

Could we setup the serializers in a separate method?

src/som/vmobjects/SBlock.java
```diff
@@ -51,6 +51,10 @@ public MaterializedFrame getContext() {
     return context;
   }
 
+  public MaterializedFrame getContextOrNull() {
```
smarr, line 54: 👍

src/som/vmobjects/SClass.java
```diff
@@ -66,6 +72,8 @@
 
   @CompilationFinal private ClassFactory instanceClassGroup; // the factory for this object
 
+  @CompilationFinal protected AbstractSerializationNode serializer;
```
smarr: This is a node that specializes itself? I am not sure this is going to work if any `rewrite` is going to be used.
But I only saw specializations, so, might be ok.

src/som/vmobjects/SClass.java
```diff
@@ -341,4 +360,12 @@ public String toString() {
   public MaterializedFrame getContext() {
     return context;
   }
+
+  public AbstractSerializationNode getSerializer() {
+    return serializer;
+  }
+
+  public void setSerializer(final AbstractSerializationNode serializer) {
```
smarr: Is this setter strictly necessary? Can't this be done via the constructor?
Then the field could be final, too.

src/som/vmobjects/SInvokable.java
```diff
@@ -65,6 +68,9 @@ public SInvokable(final SSymbol signature,
     this.invokable = invokable;
     this.callTarget = invokable.createCallTarget();
     this.embeddedBlocks = embeddedBlocks;
+    if (VmSettings.TRACK_SNAPSHOT_ENTITIES) {
```
smarr: Is this necessary?

We have the `StructuralProbe` mechanism in the parser.
Could that be used instead?

Structural probes have `recordNewMethod(SInvokable method)`.

src/som/vmobjects/SSymbol.java
```diff
@@ -40,9 +41,13 @@
   public SSymbol(final String value) {
     string = value;
     numberOfSignatureArguments = determineNumberOfSignatureArguments();
-    if (VmSettings.KOMPOS_TRACING) {
+    if (VmSettings.KOMPOS_TRACING || VmSettings.ACTOR_TRACING
+        || VmSettings.TRACK_SNAPSHOT_ENTITIES) {
       symbolId = (short) idGenerator.getAndIncrement();
       TracingBackend.logSymbol(this);
+      if (VmSettings.TRACK_SNAPSHOT_ENTITIES) {
+        SnapshotBackend.registerSymbol(this);
```
smarr, line 49: Hmm, I am not a huge fan that we have now two things doing the same thing roughly.

Can this be avoided?

src/tools/snapshot/SnapshotBackend.java
```diff
@@ -0,0 +1,76 @@
...
+import som.vm.VmSettings;
+import som.vmobjects.SClass;
+import som.vmobjects.SInvokable;
+import som.vmobjects.SSymbol;
+
+
+public class SnapshotBackend {
+  private static byte snapshotVersion = 0;
+
+  private static EconomicMap<Short, SSymbol>      symbolDictionary;
```
smarr: should these be final?

src/tools/snapshot/SnapshotBuffer.java
```diff
@@ -0,0 +1,73 @@
...
+  public static final int MAX_FIELD_CNT = Byte.MAX_VALUE;
+
+  public SnapshotBuffer() {
+    super(true);
+    this.entries = EconomicMap.create();
+  }
+
+  // This map allows us to know if we already serialized an object (and avoid circles)
+  // We can get the location of the serialized object in the trace
+  private EconomicMap<Object, Long> entries;
```
smarr: fields are defined before the constructor.

Should this be final?

src/tools/snapshot/SnapshotBuffer.java
```diff
@@ -0,0 +1,73 @@
...
+
+    this.putShortAt(this.position,
+        Symbols.symbolFor(cls.getMixinDefinition().getIdentifier()).getSymbolId());
+    this.position += CLASS_ID_SIZE + (FIELD_SIZE * fieldCnt);
+    return oldPos + CLASS_ID_SIZE;
+  }
+
+  @Override
+  protected void swapBufferWhenNotEnoughSpace(final TraceActorContextNode tracer) {
+    throw new UnsupportedOperationException();
```
smarr: `UnsupportedOperationException(msg)` takes a string. Would be better to put what you have in the comment as a string there.

src/tools/snapshot/nodes/ObjectSerializationNodes.java
```diff
@@ -0,0 +1,274 @@
...
+
+  @GenerateNodeFactory
+  public abstract static class SObjectSerializationNode
+      extends ObjectSerializationNode {
+
+    protected final int                         fieldCnt;
+    @Children private CachedSlotRead[]          fieldReads;
+    @Children private CachedSlotWrite[]         fieldWrites;
+    @Children private CachedSerializationNode[] cachedSerializers;
+    protected ObjectLayout                      layout;
```
smarr: should this be final?

src/tools/snapshot/nodes/ObjectSerializationNodes.java
```diff
@@ -0,0 +1,274 @@
...
+      }
+
+      if (!layout.isValid()) {
+        // deoptimize and adopt read nodes that fit the current layout
+        CompilerDirectives.transferToInterpreterAndInvalidate();
+        layout = clazz.getLayoutForInstancesUnsafe();
+        CachedSlotRead[] newChildren = createReadNodes(so);
+
+        for (int i = 0; i < fieldCnt; i++) {
+          fieldReads[i].replace(newChildren[i]);
```
smarr: this looks brittle and I can't tell whether this even works.

I was wondering whether simply replacing the full node (`this`) would be saver?

smarr (commented):
I think generally fine, there are some cosmetics in here, I'd like you to look at.

Otherwise, I am happy to merge it without anything extra. Though, I doubt this code will survive the first benchmark unchanged.

src/som/vmobjects/SInvokable.java
daumayr: SInvokable information could be used from the probes if I can change it to a map. For the SClasses it looks different, I don't see a connection from tracked mixins to the SClasses.

smarr: ?

There is no connection between mixins and classes. Mixins are class bodies. Classes are a mixin + a super class + mixins + an identity.
A class group is a mixin + a super class + mixins but without identity.

core-lib/TestSuite/Serialization.ns
```diff
@@ -0,0 +1,186 @@
...
+  private TransferArray = platform kernel TransferArray.
+  private ValueArray = platform kernel ValueArray.
+  private Vector = platform kernel Vector.
+|)(
+
+  public class ValueClass new: a and: b = Value (
+  | public f1 = a.
+    public f2 = b.|
+  )(
+    public verify: clone =(
```
smarr: This isn't used, is it? Do we need it?
`con` is not defined.

core-lib/TestSuite/Serialization.ns
```diff
@@ -0,0 +1,186 @@
...
+      |clone|
+      clone:: Actors snapshotClone: obj.
+      assert: obj equals: clone.
+    )
+
+    public cloneAndVerifyArray: arr = (
+      | clone |
+      clone:: Actors snapshotClone: arr.
+      arr doIndexes: [ :i |
+        assert: (arr at: i) equals: (clone at: i).
```
smarr, line 82: method name says clone and verify, but does clone and equals

core-lib/TestSuite/Serialization.ns
```diff
@@ -0,0 +1,186 @@
...
+      public verify: clone =(
+        ^(((((f1 = clone f1) and:
+        (f1 = clone f1)) and:
+        (f1 = clone f1)) and:
+        (f1 = clone f1)) and:
+        (f1 = clone f1)) and:
+        (f1 = clone f1)
+      )
+    )
+
```
smarr, line 65: only one empty line please

core-lib/TestSuite/Serialization.ns
```diff
@@ -0,0 +1,186 @@
...
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*)
+class SerializationTests usingPlatform: platform testFramework: minitest = Value(
+| private TestContext = minitest TestContext.
+  private Actors = platform actors.
```
smarr: This is a module instance, please do not capitalize.
Only classes are capitalize.

core-lib/TestSuite/Serialization.ns
```diff
@@ -0,0 +1,186 @@
...
+  private String      = platform kernel String.
```
smarr: Why the strange white space here? Inconsistent with the rest.

core-lib/TestSuite/Serialization.ns
```diff
@@ -0,0 +1,186 @@
...
+        (f1 = clone f1)) and:
+        (f1 = clone f1)) and:
+        (f1 = clone f1)
+      )
+    )
+
+
+
+    public cloneAndVerify: obj = (
+      |clone|
```
smarr: please `| clone |`, i.e. with spaces

core-lib/TestSuite/Serialization.ns
```diff
@@ -0,0 +1,186 @@
...
+      clone:: Actors snapshotClone: obj.
+      assert: (obj verify: clone).
+    )
+
+    public cloneAndEquals: obj = (
+      |clone|
```
smarr: please `| clone |`, i.e. with spaces

core-lib/TestSuite/Serialization.ns
```diff
@@ -0,0 +1,186 @@
...
+      cloneAndEquals: 754235.
+      cloneAndEquals: -12312684.
+      cloneAndEquals: 234.64563.
+      cloneAndEquals: nil.
+      cloneAndEquals: 'testString123'.
+      cloneAndEquals: ''.
+    )
+
+    public testObjectWithoutFields = (
+      |original clone |
```
smarr: missing space between pipe and `original`

src/som/compiler/MixinDefinition.java
smarr: @daumayr what's with this?

src/som/primitives/SystemPrims.java
smarr: @daumayr ?

src/som/primitives/SystemPrims.java
smarr: @daumayr ?

src/som/primitives/SystemPrims.java
smarr: @daumayr ?

src/som/vmobjects/SClass.java
```diff
@@ -66,26 +76,50 @@
 
   @CompilationFinal private ClassFactory instanceClassGroup; // the factory for this object
 
+  @CompilationFinal protected SerializerRootNode serializationRoot;
```
smarr: This should be final now, no?

src/som/vmobjects/SClass.java
```diff
@@ -341,4 +386,12 @@ public String toString() {
   public MaterializedFrame getContext() {
     return context;
   }
+
+  public void serialize(final Object o, final SnapshotBuffer sb) {
+    serializationRoot.getSerializer().execute(o, sb);
+  }
+
+  public AbstractSerializationNode getSerializer() {
+    return ((SerializerRootNode) serializationRoot.getRootNode()).getSerializer();
```
smarr: This should be `serializationRoot.getSerializer()` no?

smarr (changes requested):
Here a bunch of other cosmetic issues.

And there are still other things you haven't respondent to, or changed.
Would be good if we could at least discuss them.

src/som/compiler/MixinDefinition.java
daumayr: sorry, overlooked this comment. Do you think we should lazily cache the identifiers (e.g. as symbol) ?

smarr: That's not what my comment asks. I am more interested in avoiding the duplicated code.

daumayr: Yes, I can do that.
If we want the toString of ClassFactory to remain the same when using the new method in MixinDefinition, I have to drop the path portion. Dropping the path would be ok as long as Module names are unique.

src/som/compiler/MixinDefinition.java
daumayr: currently absolute but I can make it relative

src/som/compiler/MixinDefinition.java
smarr: There is no requirement for toString to remain the same, I think. Don't think that module names can be assumed unique. There's nothing guaranteeing this, unfortunately.

src/som/compiler/MixinDefinition.java
```diff
@@ -851,12 +853,20 @@ public static boolean isInstance(final TruffleObject obj) {
...
+   * @return the fully qualified name of this MixinDefinition
+   */
   public String getIdentifier() {
     MixinDefinition outer = getOuterMixinDefinition();
     if (outer != null) {
       return outer.getIdentifier() + "." + this.name.getString();
     } else if (this.isModule && this.sourceSection != null) {
-      return this.sourceSection.getSource().getPath() + ":" + this.name.getString();
+      Path absolute = Paths.get(this.sourceSection.getSource().getURI());
+      Path relative = Paths.get("").toAbsolutePath().relativize(absolute);
```
smarr: This is going to change when the working directory is changed by the program, right?

src/som/vmobjects/SClass.java
```diff
@@ -89,7 +90,12 @@ public SClass(final SObjectWithClass enclosing,
       final NodeFactory<? extends AbstractSerializationNode> factory) {
     this.enclosingObject = enclosing;
     this.context = null;
-    this.serializationRoot = new SerializerRootNode(null, factory.createNode(this));
+    if (VmSettings.SNAPSHOTS_ENABLED) {
+      CompilerDirectives.transferToInterpreter();
+      this.serializationRoot = new SerializerRootNode(null, factory.createNode(this));
```
smarr: The `transferToInterpreter` makes me wonder whether the `serializationRoot` shouldn't go into the `instanceClassGroup`.
Also, please avoid the generic `factory` name for the parameter, it's confusing in this context.

src/som/compiler/MixinDefinition.java
daumayr: Right, do we have the SOMns path stored somewhere?

smarr: Not really, I think. There are various things passed in as `-D` flags or parameter in the `som` launcher script. For instance the kernel file, platform file, and tools directory. Perhaps we want the core-lib folder as relative root, too?

src/som/vmobjects/SClass.java
daumayr: Moving the serializationRoot shouldn't be too much work. Do you wan't me to move it, or add a comment/TODO and consider it later.

src/som/compiler/MixinDefinition.java
daumayr: We don't need a new parameter for this, we can just use the directory containing the kernel file.

src/som/vmobjects/SClass.java
smarr: Conceptually it seems to belong into `instanceClassGroup`. `SClass` should really just provide identity. Everything else should be in the more stable structures. We don't want a separate serializer per instance of `SClass` either, I think.

daumayr: From what i saw ObjectLiterals create a new SClass for each instance. Moving the serializer would be beneficial in that case, and ensure there is only one.

smarr: Not just object literals, any nested class is potentially instantiated many times.
Inner classes for different outer objects do always have separate identities.

src/som/interpreter/objectstorage/ClassFactory.java
```diff
@@ -80,6 +92,14 @@ public ClassFactory(final SSymbol name, final MixinDefinition mixinDef,
 
     this.superclassAndMixins = superclassAndMixins;
 
+    if (VmSettings.SNAPSHOTS_ENABLED) {
+      CompilerDirectives.transferToInterpreter();
```
smarr: This `transferToInterpreter()` shouldn't be done here, because it interferes with the  `VM.callerNeedsToBeOptimized` a few lines below. Perhaps initializing `serializationRoot` should come afterwards. Either way, we expect this code to be executed only on the slow path.

src/som/interpreter/objectstorage/ClassFactory.java
```diff
@@ -174,4 +202,13 @@ public String toString() {
     }
     return "ClsFct[" + mixinDef.getIdentifier() + s + "]";
   }
+
+  public SSymbol getIdentifier() {
+    if (identifier == null) {
+      CompilerDirectives.transferToInterpreterAndInvalidate();
+      identifier = Symbols.symbolFor(mixinDef.getIdentifier());
```
smarr: Hm, why is the caching logic here and not in the mixin definition?

smarr (approved):

PR #276, opened 2018-10-23
Serialization Improvements

daumayr:
This PR is another step towards actor snapshots, it introduces:

- **Support for deserializing circular object graphs** by introducing a fixup system:
When a not yet available object is referenced, we temporarily assign a placeholder (e.g. `null`), and create a `FixupInformation` object that allows us to insert the correct reference once the object is available (deserialized).

- **Serialization of blocks:**
Blocks can be (de)serialized, including their context (`MaterializedFrame`).
We do not serialize `FrameDescriptors` and instead recover that information in replay from the blocks invokable.
Non-local returns from deserialized blocks are currently not supported (this means we cover sufficient semantics of for instance JS).

- **Serialization of Promises and Messages:**
Promises including any scheduled callbacks, messages, or chained promises can be (de)serialized.
Message serialization will be essential for snapshotting, as it will be our starting point for snapshots (i.e. serializing all objects that are reachable from serialized messages).

### DIFF
--- a/core-lib/TestSuite/Serialization.ns
+++ b/core-lib/TestSuite/Serialization.ns
@@ -129,15 +129,16 @@ class SerializationTests usingPlatform: platform testFramework: minitest = Value
     )
 
     (*
-    not context free... context also contains a reference to the block itself,
-    such cyclic graphs are currently not surrported
-    public testBlockWithoutContext = (
+    not context free...*)
+    public testBlock = (
       | original clone |
-      original:: blockWC.
+      original:: [ :val :val2 |
+          val * val2.
+      ].
       clone:: actors snapshotClone: original.
-      assert: (original value: 21 with: 2) equals: (clone value: 21 with: 2).
+      assert: (clone value: 21 with: 2) equals: (original value: 21 with: 2).
     )
-    *)
+
 
     public testSimpleArrays = (
       cloneAndEqualsArray: { 1 . 20 . 33 . 404 }.
@@ -178,11 +179,10 @@ class SerializationTests usingPlatform: platform testFramework: minitest = Value
       pp:: actors createPromisePair.
       p:: pp promise.
       pc:: actors snapshotClone: p.
-
     )
 
-    public testCloneNotResolved = (
-      | pp p pc|
+    public testAsyncCloneNotResolved = (
+      | pp p pc rp|
       pp:: actors createPromisePair.
       p:: pp promise.
       pc:: actors snapshotClone: p.
@@ -190,7 +190,7 @@ class SerializationTests usingPlatform: platform testFramework: minitest = Value
         assert: false.
       ].
       pp resolve: 1.
-      assert: pp promise resolvedWith: 1.
+      ^ (assert: pp promise resolvedWith: 1).
     )
 
     public testSimpleResolver = (
@@ -202,38 +202,64 @@ class SerializationTests usingPlatform: platform testFramework: minitest = Value
       r resolve: 2.
     )
 
-    public testSimplePromisePair = (
-      | pp ppc|
+    public testAsyncSimplePromisePair = (
+      | pp ppc |
       pp:: actors createPromisePair.
       ppc:: actors snapshotClone: pp.
       pp resolve: 42.
-      assert: pp promise resolvedWith: 42.
       ppc resolve: 21.
-      assert: ppc promise resolvedWith: 21.
+      ^ (assert: pp promise resolvedWith: 42),
+      (assert: ppc promise resolvedWith: 21).
     )
 
-    public testCloneResolved = (
-      | pp p pc|
+    public testAsyncCloneResolved = (
+      | pp p pc rp rp2|
       pp:: actors createPromisePair.
       pp resolve: 84.
       p:: pp promise.
       pc:: actors snapshotClone: p.
-      assert: p resolvedWith: 84.
-      assert: pc resolvedWith: 84.
+      ^ (assert: p resolvedWith: 84),
+      (assert: pc resolvedWith: 84).
     )
 
-    public testClonePromiseWithMessage = (
+     public testAsyncCloneChain = (
+      | pp pp2 pp3 ppc|
+      pp:: actors createPromisePair.
+      pp2:: actors createPromisePair.
+      pp3:: actors createPromisePair.
+      pp2 resolve: pp.
+      pp3 resolve: pp2.
+      ppc:: actors snapshotClone: pp.
+
+      ppc resolve: 42.
+      ^ assert: ppc promise resolvedWith: 42.
+      (*TODO check that pp2 and three were cloned...*)
+    )
+
+    public testAsyncClonePromiseWithMessage = (
       | pp ppc|
       pp:: actors createPromisePair.
       pp promise <-: value: 42.
       ppc:: actors snapshotClone: pp.
       ppc resolve: [ :v |
-        v println.
       ].
       pp resolve: [ :v |
-        v println.
-      ]
+      ].
+      ^ pp promise, ppc promise.
     )
+
+    public testAsyncCloneCallback = (
+      | pp ppc rp |
+      pp:: actors createPromisePair.
+      pp promise whenResolved: [ :val |
+        assert: val equals: 42.
+      ].
+      ppc:: actors snapshotClone: pp.
+      pp resolve: 42.
+      ppc resolve: 42.
+      ^ (assert: pp promise resolvedWith: 42),
+      (assert: ppc promise resolvedWith: 42).
+  )
 
   ) : ( TEST_CONTEXT = () )
 )
--- a/core-lib/TestSuite/Serialization.ns
+++ b/core-lib/TestSuite/Serialization.ns
@@ -20,6 +20,7 @@ THE SOFTWARE.
 *)
 class SerializationTests usingPlatform: platform testFramework: minitest = Value(
 | private TestContext = minitest TestContext.
+  private AsyncTestContext = minitest AsyncTestContext.
   private actors = platform actors.
   private String = platform kernel String.
   private Array = platform kernel Array.
@@ -33,11 +34,7 @@ class SerializationTests usingPlatform: platform testFramework: minitest = Value
     public f2 = b.|
   )()
 
-  public class SerializationTest = TestContext (
-  | blockWC = [ :a :b |
-      ^ a * b.
-    ].
-  |)(
+  public class SerializationTest = AsyncTestContext ()(
 
     public class EmptyClass = ()(
       public test = (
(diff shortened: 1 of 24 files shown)


core-lib/TestSuite/Minitest.ns
```diff
@@ -441,12 +441,12 @@ OTHER DEALINGS IN THE SOFTWARE. *)
...
-                  ifNil:    [ TestFailure case: self description: ex messageText]
-                  ifNotNil: [:it | it createFailureResultFor: self description: ex messageText])
+                  ifNil:    [ TestFailure case: self description: e messageText]
+                  ifNotNil: [:it | it createFailureResultFor: self description: e messageText])
               ] ifFalse: [
                 promisePair resolve: (testContextInstance
-                  ifNil:    [ TestError case: self exception: ex ]
-                  ifNotNil: [:it | it createErrorResultFor: self exception: ex ])
+                  ifNil:    [ TestError case: self exception: e ]
+                  ifNotNil: [:it | it createErrorResultFor: self exception: e ])
```
smarr: Why did you do the renamings in this file? Seems unnecessary, and unrelated to what you want to achieve in this PR.

core-lib/TestSuite/Minitest.ns
```diff
@@ -755,12 +755,28 @@ OTHER DEALINGS IN THE SOFTWARE. *)
...
+        whenResolved: [:actualResolution |
+          resolved ifFalse: [
+            resolved:: true.
+            assert: actualResolution equals: expectedResolution.
+            assertPP resolve: 1.
+          ].
+        ]
+        onError: [:actualResolution |
+          failWithMessage: 'Promise broken with error: ' + actualResolution asString ].
+      ^ assertPP promise.
```
smarr: What are all theses changes about?
This looks problematic to me.
And it doesn't seem to be a self-contained and documented change either.
The time out is rather arbitrary, too.

If this is needed, or desired, could you please break this change out?
And would be good to have an `assert:resolvedWith:timeout:` which takes an explicit parameter. 
`assert:resolvedWith:` can then call it explicitly.

core-lib/TestSuite/Serialization.ns
```diff
@@ -20,6 +20,7 @@ THE SOFTWARE.
 *)
 class SerializationTests usingPlatform: platform testFramework: minitest = Value(
 | private TestContext = minitest TestContext.
```
smarr: Do you still need the `TestContext`?

src/som/interpreter/actors/EventualMessage.java
```diff
@@ -203,10 +215,10 @@ public boolean getHaltOnPromiseMessageResolution() {
...
    */
   public abstract static class AbstractPromiseSendMessage extends PromiseMessage {
-    private final SSymbol    selector;
-    protected Actor          target;
-    protected Actor          finalSender;
-    protected final SPromise originalTarget;
+    private final SSymbol selector;
+    protected Actor       target;
+    protected Actor       finalSender;
+    protected SPromise    originalTarget;
```
smarr: Could you at least make it `@CompilationFinal` then? This is likely a performance regression.

src/som/interpreter/actors/EventualMessage.java
```diff
@@ -281,7 +308,7 @@ protected PromiseSendMessage(final SSymbol selector, final Object[] arguments,
     /**
      * The promise on which this callback is registered on.
      */
-    protected final SPromise promise;
+    protected SPromise promise;
```
smarr: Could you at least make it `@CompilationFinal` then? This is likely a performance regression.

src/tools/snapshot/nodes/ObjectSerializationNodes.java
```diff
@@ -234,12 +235,32 @@ public Object deserialize(final ByteBuffer sb) {
...
+      public SlotFixup(final SObject obj, final CachedSlotWrite csw) {
+        this.csw = csw;
+        this.obj = obj;
+      }
+
+      @Override
+      public void fixUp(final Object res) {
+        csw.doWrite(obj, res);
+      }
+
```
smarr:
```suggestion
```

src/som/interpreter/actors/SPromise.java
```diff
@@ -37,6 +40,15 @@ public static SPromise createPromise(final Actor owner,
     }
   }
```
smarr, line 42:
```suggestion

/** Used by snapshot deserialization. */
```

src/som/interpreter/actors/SPromise.java
```diff
@@ -94,6 +106,29 @@ public final boolean isValue() {
     return false;
   }
 
+  /**
+   * Do not use for things other than serializing Promises.
+   *
+   * @return the value this promise was resolved to
+   */
+  public final Object getValue() {
```
smarr:
```suggestion
  public final Object getValueForSnapshot() {
```

I would try to be as explicit as possible.
Also, there is already an `getValueUnsync()` method, which would make me assume that you likely need to have a `synchronized` in here. You have race condition here at the moment.

src/som/interpreter/actors/SPromise.java
```diff
@@ -94,6 +106,29 @@ public final boolean isValue() {
...
+    assert resolutionState == Resolution.SUCCESSFUL || resolutionState == Resolution.ERRONEOUS;
+    assert value != null;
+    return value;
+  }
+
+  /**
+   * Do not use for things other than deserializing Promises.
+   * This is necessary for circular object graphs.
+   */
+  public final void setValue(final Object value) {
```
smarr:
```suggestion
  public final void setValueFromSnapshot(final Object value) {
```

Or something along those lines would be better here, too.

src/som/interpreter/actors/SPromise.java
```diff
@@ -230,6 +274,72 @@ final Object getValueUnsync() {
     return value;
   }
 
+  public int getNumChainedPromises() {
```
smarr: Everything here in this changed block is inherently racy.

Especially, if we do not do a stop the world, then this is going to break if we do not synchronize this properly somehow.

src/som/vm/ObjectSystem.java
```diff
@@ -402,6 +410,10 @@ public SObjectWithoutFields initialize() {
     Classes.blockClass.getSOMClass()
                       .setClassGroup(Classes.metaclassClass.getInstanceFactory());
 
+    setDummyClassFactory(Classes.messageClass, MessageSerializationNodeFactory.getInstance());
```
smarr, line 415:
```suggestion
    // these classes are not exposed in Newspeak directly, and thus, do not yet have a class factory
    setDummyClassFactory(Classes.messageClass, MessageSerializationNodeFactory.getInstance());
```

src/som/vm/constants/Classes.java
```diff
@@ -60,5 +63,8 @@
     falseClass = ObjectSystem.newEmptyClassWithItsClass("False");
 
     blockClass = ObjectSystem.newEmptyClassWithItsClass("Block");
+
+    messageClass = ObjectSystem.newEmptyClassWithItsClass("Message");
+
```
smarr:
```suggestion
```

src/tools/snapshot/SnapshotBackend.java
```diff
@@ -74,6 +76,11 @@ public static byte getSnapshotVersion() {
     return snapshotVersion;
   }
 
+  public static Actor lookupActor(final int actorId) {
+    // TODO implement
```
smarr: What's this todo about?

src/tools/snapshot/deserialization/DeserializationBuffer.java
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  public int position() {
+    return wrapped.position();
+  }
+
+  public void position(final int newPosition) {
+    wrapped.position(newPosition);
+  }
+
```
smarr:
```suggestion
```

src/tools/snapshot/nodes/MessageSerializationNode.java
```diff
@@ -0,0 +1,454 @@
...
+            Types.getClassOf(args[i]).serialize(args[i], sb);
+          }
+          sb.putLongAt((base + 1) + i * Long.BYTES, sb.getObjectPointer(args[i]));
+        }
+      }
+    }
+  }
+
+  /**
+   *
```
smarr:
```suggestion
```

src/tools/snapshot/nodes/MessageSerializationNode.java
```diff
@@ -0,0 +1,454 @@
...
+    base += COMMONALITY_BYTES;
+
+    doArguments(args, base, sb);
+
+    return start;
+  }
+
+  @Override
+  public EventualMessage deserialize(final DeserializationBuffer bb) {
+
```
smarr:
```suggestion
```

src/tools/snapshot/nodes/MessageSerializationNode.java
```diff
@@ -0,0 +1,454 @@
...
+        (TracingActor) dm.getSender(), base, sb);
+    base += COMMONALITY_BYTES;
+
+    doArguments(args, base, sb);
+
+    return start;
+  }
+
+  @Override
+  public EventualMessage deserialize(final DeserializationBuffer bb) {
```
smarr, line 293: I am really not a fan of such large methods.
Can't you break the cases into nicely named methods?

src/tools/snapshot/nodes/PromiseSerializationNodes.java
```diff
@@ -0,0 +1,202 @@
...
+
+      Object value = prom.getValue();
+      if (!sb.containsObject(value)) {
+        Types.getClassOf(value).serialize(value, sb);
+      }
+      sb.putLongAt(base + 1, sb.getObjectPointer(value));
+    }
+
+    @Specialization(guards = "!prom.isCompleted()")
+    public void doUnresolved(final SPromise prom, final SnapshotBuffer sb) {
```
smarr, line 53: Again, a looong method.
Can't you break the steps into smaller methods with nice names?

src/tools/snapshot/nodes/PromiseSerializationNodes.java
```diff
@@ -0,0 +1,202 @@
...
+            p = cpe.get(i);
+            SPromise.getPromiseClass().serialize(p, sb);
+            sb.putLongAt(base + i * Long.BYTES, sb.getObjectPointer(p));
+          }
+        }
+      }
+    }
+
+    @Override
+    public SPromise deserialize(final DeserializationBuffer sb) {
```
smarr, line 158: And another one.

src/tools/snapshot/nodes/PromiseSerializationNodes.java
```diff
@@ -0,0 +1,202 @@
...
+      }
+
+      @Override
+      public void fixUp(final Object o) {
+        promise.setValue(o);
+      }
+    }
+  }
+
+  // Resolvers are values and can be passed directly without being wrapped
```
smarr: Should this be a JavaDoc comment on the class?

src/tools/snapshot/nodes/PromiseSerializationNodes.java
```diff
@@ -0,0 +1,202 @@
...
+  // From what i have seen there is only one resolver created for any promise.
+  // Resolver contains a reference to the promise, which knows it's owner.
+  // If Identity of Resolvers becomes an issue just add the actor information and turn into a
+  // singleton when deserializing
```
smarr: There is only a single resolver for every promise. But promises may be chained.

smarr (changes requested):
Thanks for the PR.

And yes, this was really necessary, I think.
We are well above 1000 LOC change again.

This all is rather more complex than I had hoped :(

core-lib/TestSuite/Minitest.ns
daumayr: There is no ex variable in that scope, only e. Instead of fixing the usage I can rename e to ex if you prefer.

core-lib/TestSuite/Minitest.ns
daumayr: The current `assert:resolvedWith:` ignores the case where the promise is not resolved. I'll remove the change from this PR and put it into a Separate one with a version for specifying timeout.

core-lib/TestSuite/Minitest.ns
smarr: Ahhhh. No, the change is good. My comment just shows that it would help me to get more context for such changes (possibly by separate maintenance PRs, added tests, ...)

I mean, it's already good that you have a separate commit for the timeout stuff. But it's unfortunate that it doesn't come with any testing, any any comment for the rational for the changes.

core-lib/TestSuite/Minitest.ns
smarr: Thank you.

smarr: Just for an explanation: I didn't actually notice that this was in a separate commit. Missed that yesterday.

src/som/compiler/Parser.java
```diff
@@ -1238,6 +1238,9 @@ private ExpressionNode primary(final MethodBuilder builder) throws ProgramDefini
         SInvokable blockMethod = bgenc.assemble(blockBody,
             AccessModifier.BLOCK_METHOD, lastMethodsSourceSection);
         builder.addEmbeddedBlockMethod(blockMethod);
+        if (structuralProbe != null && VmSettings.TRACK_SNAPSHOT_ENTITIES) {
```
daumayr: probably should switch order of conditions (not sure if this gets optimized)

src/som/interpreter/actors/SPromise.java
daumayr: Probably would be best to perform all the gets in one synchronised block in the PromiseSerializationNode, and add Unsynced to the method names.

smarr (approved):
Thanks for the changes! looks good to me!

smarr:
Going to merge once all tests completed, and the benchmarks ran.

PR #278, opened 2018-11-16
Snapshot writing

daumayr:
This PR introduces snapshot creation by serialising received messages that stem from before the snapshot. Each processing thread holds a SnapshotBuffer, similar to TracingBuffers.
The most recent snapshot is persisted on VM shutdown.
Also introduces splitting of the trace (and external data) in separate files for each snapshot.


build.xml
```diff
@@ -409,6 +409,11 @@
       </exec>
     </target>
 
+    <target name="snapshot-tests" depends="compile">
```
smarr, line 412: Hm, are these tests executed on Travis?
Doesn't look like it to me.

src/som/interpreter/actors/Actor.java
```diff
@@ -275,10 +277,24 @@ protected void processCurrentMessages(final ActorProcessingThread currentThread,
         final WebDebugger dbg) {
       assert size > 0;
 
+      if (VmSettings.SNAPSHOTS_ENABLED && !VmSettings.TEST_SNAPSHOTS) {
+        SnapshotBuffer sb = currentThread.getSnapshotBuffer();
+
+        if (SnapshotBackend.getSnapshotVersion() > sb.getSnapshotVersion()) {
```
smarr: Is this check and the selection of a new `sb` something that should be done as part of the getter method?
This is mostly about responsibilities.
Seems that it's a little brittle if the client has to do the check.

src/som/interpreter/actors/EventualMessage.java
```diff
@@ -14,14 +14,22 @@
...
 
 public abstract class EventualMessage {
   protected final Object[]       args;
   protected final SResolver      resolver;
   protected final RootCallTarget onReceive;
-  protected final long           messageId;
+
+  /**
+   * Contains the messageId for Kompos tracing.
+   * This field is reused for snapshotting, where it contains the the snapshot version at send
```
smarr:
```suggestion
   * This field is reused for snapshotting. It then contains the snapshot version at send
```

src/som/interpreter/actors/EventualMessage.java
```diff
@@ -73,8 +81,15 @@ public SourceSection getTargetSourceSection() {
 
   public long serialize(final SnapshotBuffer sb) {
     ReceivedRootNode rm = (ReceivedRootNode) this.onReceive.getRootNode();
-    // Not sure if this is optimized, worst case need to duplicate this for all messages
-    return rm.getSerializer().execute(this, sb);
+
+    if (VmSettings.TEST_SNAPSHOTS || VmSettings.TEST_SERIALIZE_ALL
+        || sb.getSnapshotVersion() > this.getMessageId()) {
```
smarr:
```suggestion
        || sb.needsToBeSnapshot(getMessageId())) {
```

This would be one way to encapsulate the logic.

I am also rather surprised by all the `||` in the expression. Shouldn't it be first an `&&` and then in parentheses a `||`? I am just guess, but, an all `||` looks odd.

src/tools/concurrency/TracingActivityThread.java
```diff
@@ -78,9 +83,13 @@ public synchronized boolean swapTracingBufferIfThreadSuspendedInDebugger() {
...
   public TracingActivityThread(final ForkJoinPool pool) {
     super(pool);
+    if (VmSettings.SNAPSHOTS_ENABLED) {
+      this.snapshotBuffer = new SnapshotBuffer((ActorProcessingThread) this);
+    }
+
     if (VmSettings.ACTOR_TRACING || VmSettings.KOMPOS_TRACING) {
       threadId = threadIdGen.getAndIncrement();
-      traceBuffer = TraceBuffer.create(threadId);
+      traceBuffer = TraceBuffer.create(this.threadId);
```
smarr: why does the `this.` here need to be made explicit?

src/tools/concurrency/TracingBackend.java
```diff
@@ -252,7 +264,9 @@ public static final void forceSwapBuffers() {
       }
     }
 
-    while (!fullBuffers.isEmpty()) {
+    // forceSwap should only happen for kompos.
```
smarr, line 281: There assertion at the top of the method isn't exactly clear on that.
In Kompos, we force swap to be able to get results with good latency.

On shutdown, we probably also force swap, no? Which would be independent from Kompos.

src/tools/concurrency/TracingBackend.java
```diff
@@ -79,21 +80,16 @@
...
   // contains symbols that need to be written to file/sent to debugger,
   // e.g. actor type, message type
   private static final ArrayList<SSymbol> symbolsToWrite =
       new ArrayList<>();
 
   private static FrontendConnector front = null;
 
-  private static long              collectedMemory = 0;
+  private static long collectedMemory = 0;
   private static TraceWorkerThread workerThread;
```
smarr: Please distinguish the purposes of these two worker threads. It's unclear why we need two. Numbering doesn't help.

src/tools/concurrency/TracingBackend.java
```diff
@@ -286,6 +306,33 @@ public static void startTracingBackend() {
     workerThread.start();
   }
 
+  private static TraceWorkerThread getResponsibleWorker(final byte snapshotVersion) {
+    if (snapshotVersion % 2 == 0) {
```
smarr: What's the purpose of this?

src/tools/concurrency/TracingBackend.java
```diff
@@ -286,6 +306,33 @@ public static void startTracingBackend() {
...
+      return workerThread;
+    } else {
+      return workerThread2;
+    }
+  }
+
+  @TruffleBoundary
+  public static void switchTrace(final byte newSnapshotVersion) {
```
smarr: And this?

I don't really know what the background to this is.

src/tools/concurrency/TracingBackend.java
```diff
@@ -335,8 +402,8 @@ private BufferAndLimit tryToObtainBuffer() {
 
     private void writeExternalData(final FileOutputStream edfos) throws IOException {
       while (!externalData.isEmpty()) {
-        Object[] o = externalData.poll();
-        for (Object oo : o) {
+        Object[] ed = externalData.poll();
```
smarr: what's `ed`? I don't understand abbreviations!

src/tools/concurrency/TracingBackend.java
```diff
@@ -489,10 +556,14 @@ private void processTraceData(final FileOutputStream traceDataStream,
     }
 
     @Override
+    @TruffleBoundary
```
smarr: Why is this necessary?
This is a `run()` method on a thread, no? This should never be reached by Truffle.

src/tools/snapshot/SnapshotBuffer.java
```diff
@@ -1,47 +1,48 @@
...
-  public static final int MAX_FIELD_CNT = Byte.MAX_VALUE;
-
-  /**
-   * This map allows us to know if we already serialized an object (and avoid circles).
-   * We can get the location of the serialized object in the trace
-   */
-  private final EconomicMap<Object, Long> entries;
+  public static final int               FIELD_SIZE    = 8;
+  public static final int               CLASS_ID_SIZE = 2;
+  public static final int               MAX_FIELD_CNT = Byte.MAX_VALUE;
```
smarr: Grouping of fields is used for formatting, so, avoiding to group static and non-static fields helps with avoiding unnecessary whitespace changes.

It also emphasizes that there is a difference and makes the different types of things stand out.

src/tools/snapshot/nodes/BlockSerializationNode.java
```diff
@@ -57,7 +57,7 @@ public void serialize(final SBlock block, final SnapshotBuffer sb) {
 
       for (int i = 0; i < args.length; i++) {
         Types.getClassOf(args[i]).serialize(args[i], sb);
-        sb.putLongAt(base + (i * Long.BYTES), sb.getObjectPointer(args[i]));
+        sb.putLongAt(base + (i * Long.BYTES), sb.getRecord().getObjectPointer(args[i]));
```
smarr: The `getRecord()` is this performance critical? Does it need to be moved out of the loop?
I mean, all the code here is unoptimized, for instance the `Types.getClassOf` isn't even going to compile.

smarr (changes requested):
First look at the changes.

src/som/interpreter/actors/EventualMessage.java
daumayr: I probably should rename TEST_SNAPSHOTS to TEST_SERIALIZATION or something. It is used for the serialization test, i.e. serialization of individual object(graphs). For that test we don't trigger normal snapshots, as a result all snapshot versions are 0 and we need the ||.

src/tools/concurrency/TracingBackend.java
daumayr: On Shutdown we have the onTermination methods in TracingActivityThread that returns the buffers normally.

src/tools/concurrency/TracingBackend.java
daumayr: I renamed them to make the purposes more clear and added a description comment. 

  /**
   * Used for writing the traces in normal actor Tracing.
   * When snapshoting, this points to the worker responsible for the trace
   * associated with the current snapshot.
   */
  private static TraceWorkerThread currentWorkerThread;
  /**
   * When snapshoting, this points to the worker that handles remaining
   * trace data belonging to the previous snapshot.
   */
  private static TraceWorkerThread previousWorkerThread;

src/tools/snapshot/nodes/BlockSerializationNode.java
daumayr: I moved it out of the loop, it's a simple getter. I'm adding a todo for optimisation, first of all we need to specialise on the different blocks, then we can cache the argument serialisation.

build.xml
daumayr: Added a Travis job

smarr:
Do I need to look into this PR again? (was holding off because of failing tests)

daumayr:
I need to look into them, I've been a bit distracted by the Compiler Summit and getting replay to work.

smarr:
Sure :) just went through my todo list, and wanted to make sure that I am not blocking anything :)

.travis.yml
```diff
@@ -69,5 +70,6 @@ script:
   - if [ "$TASK" = "kompos-tests"  ]; then nvm install 8         && ant $SG && cd tools/kompos && npm -s run verify && npm test; fi
   - if [ "$TASK" = "replay1-tests" ]; then ant $SG compile       && ./tests/replay/test.sh 1; fi
   - if [ "$TASK" = "replay2-tests" ]; then ant $SG compile       && ./tests/replay/test.sh 2; fi
+  - if [ "$TASK" = "snapshot-tests" ]; then ant $SG compile       && ./tests/snapshot/test.sh; fi
```
smarr, line 73: Thanks for adding this, could you also add the snapshot tests to the gitlab setup?

smarr (changes requested):
Looks good to me, but would be great if you could still enable testing in the GitLab config.

.travis.yml
daumayr: done

smarr (approved):

PR #282, opened 2018-12-22
Minimize the Truffle-related building

smarr:
In the current configuration, we build stuff repeatedly. By using dynamic imports, we can avoid that.

Also, as it turns out, we have now a working and fully compiled Graal compiler handy, we may as well use it.


smarr:
This change avoids about 10min of build time on Travis. :)

PR #285, opened 2019-01-14
Fix object layout races

daumayr:
This PR fixes issue #85.

The Phaser used in ObjectTransitionSafepoint used a default Phaser, which Terminates when onAdvance(...) returns true. This caused the synchronisation methods to have no effect.
The default onAdvance implementation causes the Phaser to terminate when the number of registered Threads is 0.

By overwriting the onAdvance method and returning false we can prevent Phaser from terminating and preserve the synchronisation.


src/som/interpreter/objectstorage/ObjectTransitionSafepoint.java
```diff
@@ -10,6 +10,7 @@
 import com.oracle.truffle.api.nodes.InvalidAssumptionException;
 
 import som.compiler.MixinDefinition.SlotDefinition;
+import som.vm.CustomPhaser;
```
smarr: CustomPhaser, not a great name.
How about: `SafepointPhaser`?

src/som/vmobjects/SObject.java
```diff
@@ -517,4 +517,22 @@ private void setFieldAfterLayoutChange(final SlotDefinition slot,
     StorageLocation location = getLocation(slot);
     location.write(this, value);
   }
+
+  public final boolean isObjectSlotAllocated(final SlotDefinition slot) {
+    StorageLocation loc = objectLayout.getStorageLocation(slot);
+    return loc instanceof ObjectStorageLocation;
+  }
+
+  public final synchronized void ensureSlotAllocatedToAvoidDeadlock(
```
smarr, line 521: Hmmmmmm. So, we still hold a lock when we may trigger a safepoint? Any way around that?

Seems like this is only moving the deadlock?

daumayr: this method is called after the safepoint starts, and nowhere else. the synchronized is for the case that two threads enter the safepoint to update the same object.

src/som/interpreter/objectstorage/ObjectTransitionSafepoint.java
daumayr: good idea

daumayr:
While tests are running for this branch, we still have an issue with some benchmarks. The Vacation Benchmark still deadlocks, so ideally we don't merge this PR until we fix the Join Prim and similar things.

src/som/vmobjects/SObject.java
smarr: Aaaaah, right, now I see it.

Should have looked at it in a proper IDE to start with :)
Thanks!

smarr:
Ok, will see what I can do about those. Will probably try to do them in a separate PR to see performance issues independently.

daumayr:
I think we need to exclude those benchmarks (LeeTm and Vacation) until then.

src/som/interpreter/nodes/dispatch/ClassSlotAccessNode.java
```diff
@@ -60,6 +61,11 @@ public SClass read(final SObject rcvr) {
       return (SClass) cachedValue;
     }
 
+    if (!rcvr.isObjectSlotAllocated(slotDef)) {
```
smarr: Does this test compile out? We are here at a point where we are guaranteed to have one specific layout (which our guard tested before).
We may want to test against that instead of the `rcvr` object, which may have changed in the mean time.

src/som/interpreter/objectstorage/ObjectTransitionSafepoint.java
```diff
@@ -140,12 +155,27 @@ public void writeAndGeneralizeSlot(final SObject obj, final SlotDefinition slot,
...
   private void replaceAssumptionAndWaitForSafepointEnd() {
     // update the assumption
-    synchronized (this) {
-      // might have been replaced by another thread already
-      if (!noSafePoint.isValid()) {
-        noSafePoint = create();
+    int active = transitionsInProgress.decrementAndGet();
+
+    if (active == 0) {
+      synchronized (this) {
```
smarr: Is it on purpose that you got two separate synchronization mechanisms here?

This might be an optimization I guess, but I may be simpler to stick to exactly one (`synchronized` only?), if possible?

smarr (commented):
A quick review to give at least some comments.
Sorry, didn't get to more.

src/som/interpreter/objectstorage/ObjectTransitionSafepoint.java
daumayr: are you referring to the atomic int? I suppose we can change it to a normal int, and move all operations on it inside the synchronized. A synchronized then also needs to be introduced into the waitForSafepointStart.

src/som/interpreter/nodes/dispatch/ClassSlotAccessNode.java
daumayr: not sure if this compiles away, after all the rcvr isn't final. Probably would be better to use the guards layout, and maybe an assumption that get's invalidated when the slot is allocated.

smarr: This issue is now resolved with: https://github.com/smarr/SOMns/pull/285/commits/88b3b2b9a4c918c731b0a4fe6e6c951f41a371a9

Replacing a hash lookup with a constant :)

src/som/interpreter/objectstorage/ObjectTransitionSafepoint.java
smarr: Addressed with: https://github.com/smarr/SOMns/pull/285/commits/33f924f00754b362bd1ae423d52cf3f15b8df155

This way, there is one uniform mechanism, and performance is hopefully not a major issue. Though, not really measured yet.

src/som/interpreter/objectstorage/ObjectTransitionSafepoint.java
```diff
@@ -59,6 +60,19 @@ private static Assumption create() {
   @TruffleBoundary
   public void register() {
     phaser.register();
+    try {
```
smarr: @daumayr I am not sure I am getting the full logic here.

We can have the following situations:
- register while no safepoint is done
- register while safepoint is starting
- register while safepoint is in progress

Can we distinguish the last two? only by looking at `transitionsInProgress`, no?

smarr:
@daumayr I come up with what I think is a working solution. Could you please review and possible test this?
I haven't yet thought of a good test on how to stress the safepoints.
Not sure I can induce repeated safepoints easily.
Would be good to discuss this in detail, to see whether there are any know races left, or any other issue that may make this approach problematic.

The most important bit to discuss is likely https://github.com/smarr/SOMns/pull/285/commits/0274546403810d7f522e6f1ae8a8b67b8d9f2b6e

daumayr:
`(phase & 1) == 0`
I was thinking of using something similar when i did some work on the safepoints. When no safepoint is going on and all threads deregister the phaser advances to the next phase. After that the even/odd check would no longer work. You might want to change the doArrive method to not advance phase when there are no threads left.

smarr:
@daumayr hmmmm, ok. Didn't think of that. Looking at the code: https://github.com/smarr/SOMns/pull/285/files#diff-2b3be932b680eb719ab91a02f3b6f064R137

I suppose, I could simply increment the phase in that case by 2, no?

smarr:
hmmm, well, possibly depending on in which concrete state we are. Though, we should never unregister during a safepoint. So, probably increment by 2 plus an assertion that we are never in a phase that indicates a safepoint is going on

daumayr:
Increment by two sounds good, that way we still have a phase change and keep the odd/even.

smarr:
ok, cool.

and one more todo:

- [x] add unit tests that test the basic logic for correctness, and have asserts for all the parts not covered

smarr:
@daumayr could you give this a spin with your snapshot tests?
Would also be great if you could review it and look at the added test. Any obvious test we should add?

daumayr:
Snapshot Tests running, didn't see any errors.

tests/java/som/interpreter/objectstorage/SafepointPhaserTest.java
```diff
@@ -0,0 +1,96 @@
...
+  private final ClassFactory factory;
+  private final ObjectLayout layout;
+
+  public SafepointPhaserTest() {
+    SlotDefinition slotDef = new SlotDefinition(null, AccessModifier.PUBLIC, false, null);
+    EconomicSet<SlotDefinition> instanceSlots = EconomicSet.create();
+    instanceSlots.add(slotDef);
+
+    instanceClass = new SClass(null);
+    factory = new ClassFactory(null, null, instanceSlots, null, false, false, false, null,
```
daumayr, line 30: This seems to have a wrong number of arguments on my machine.

smarr: Hm, what exactly do you mean with this?
Also, is this without any other changes? Seems to compile fine on Travis and Brussels.

Issue #286, opened 2019-01-15
All Blocking Primitives need to interact with Safepoint Logic


smarr:
For the blocking primitives, we need to decide one-by-one what to do with respect to safepoints.

There is always the potential for a deadlock, if they do not unregister before blocking.
Though, there may be performance implications that need to be considered.

This needs to be addressed so that #285 can be merged.

smarr:
Got fixed with #288.

PR #288, opened 2019-01-16
Blocking primitives cannot participate in safepoints

smarr:
Adapt all known blocking primitives to unregister before, and reregister after a safepoint.

@daumayr could you have a look and see whether this fixes your issues?

We don't have benchmarks evaluating this change, so, I think, correctness is key, and if performance becomes an issue, we need to find more specific strategies, or optimizations for special cases.


daumayr:
@smarr, this fix resolves the issues in the LeeTM and Vacation Benchmarks.

smarr:
So, changes look good?
Want me to merge this one?

daumayr:
yes please, this would reduce the number of branches and PRs

smarr:
This fixed #286.

PR #289, opened 2019-01-22
Async stack trace structure (from dev)

clementbera:
Version rebased on dev branch for SMarr review.

So far basic synchronous on asynchronous benchmark are working with the different settings enabled. With the settings disabled everything should keep working too. Still need to work on printing the stack with the different settings enabled or not.

Three additional settings:
--async_stack_traces_structure: enables the shadow stack entry, allowing to recreate the stack across asynchronous calls.
--async_stack_traces_method_cache: enables per method cache to CachedDispatchNode or LexicallyBoundDispatchNode, avoiding to create shadow stack entry if a method is called from a single dispatch node.
--async_stack_traces_inline_cache: enables per shadow stack entry creation site cache, avoiding to create repeatedly the same shadow stack entry in loops, the first few times a loop is created.

async_stack_traces_structure has to be enabled for the other ones to change anything. The other 2 settings can be enabled separately or together.


build.xml
```diff
@@ -21,11 +21,11 @@
     <property name="svm.build"   location="${svm.dir}/mxbuild/dists" />
     <property name="truffle.build" location="${truffle.dir}/mxbuild/dists" />
     <property name="somns-deps.version" value="0.3.6" />
-    <property name="checkstyle.version" value="8.11" />
+    <property name="checkstyle_checkse.version" value="8.11" />
```
smarr: Why this change?

docs/basic-setup.md
```diff
@@ -48,4 +48,4 @@ Afterwards, the simple Hello World program is executed with:
 
 ```bash
 ./som core-lib/Hello.ns
-```
+```
```
smarr: Strange change?

src/som/interpreter/Invokable.java
```diff
@@ -46,6 +46,10 @@ public final boolean isAtomic() {
     return isAtomic;
   }
 
+  public ExpressionNode getRoot() {
```
smarr: `Node.getRootNode()` isn't any good for this?

src/som/interpreter/Method.java
```diff
@@ -77,6 +80,20 @@ public boolean equals(final Object o) {
     return false;
   }
 
+  public void setNewCaller(final BackCacheCallNode caller) {
```
smarr, line 82: You are changing a `@CompilationFinal` field here, do you need to invalidate any code that may have been compiled with this assumption backed in?

src/som/interpreter/actors/ErrorPromiseNode.java
```diff
@@ -10,21 +10,23 @@
 
 
 @GenerateNodeFactory
-@Primitive(primitive = "actorsError:with:isBPResolver:isBPResolution:")
+@Primitive(primitive = "actorsError:with:entry:isBPResolver:isBPResolution:")
+
```
smarr: Why the space?
Also, you seem to change the signature of the specialization, I'd assume that primitive signature would need to change then, too?

smarr (commented):
Some very initial comments, only skimming over the first bits

build.xml
clementbera: I think this was generated by mx.

src/som/interpreter/Invokable.java
clementbera: I think Node.getRootNode answers the root node, i.e. the method or primitive, while this method  answers the sequence of expression held by the root node. So I would say no.

docs/basic-setup.md
clementbera: Yes, the change should be removed but I don't know how. I checked out from dev and it is still there.

src/som/interpreter/Method.java
clementbera: Oh, I thought @CompilationFinal was taking care of that for me. Yes code needs to be invalidated.

src/som/interpreter/actors/ErrorPromiseNode.java
clementbera: I removed the space.

I am not sure about what you are asking for the signature. The signature of AbstractPromiseResolutionNode was also changed, is that what you mean?

src/som/interpreter/Method.java
clementbera: So I removed the @CompilationFinal of this variable. The problem is in BackCacheCallNode >> multipleCaller (). When this method is called, all the optimized code using the BackCacheCallNode has to be invalidated. I am not sure on how to do that.

build.xml
smarr: No, please revert. mx doesn't know anything about our build system.

src/som/interpreter/Invokable.java
smarr: would `getBodyNode()` be an appropriate name?

src/som/interpreter/Method.java
smarr: Not sure this is the ideal example: https://github.com/smarr/SOMns/blob/dev/src/som/interpreter/objectstorage/ObjectLayout.java#L153

Assumptions are the way to invalidate code "non-locally". Any code that is compiled checking an assumption is registered for being discarded when the assumption is invalidated. In the example, there is an assumption for the memory layout of classes. It is invalidated, when the class gets a new layout assigned (which in return is going to be used by its instances).

smarr:
@clementbera thanks for all the work so far. Before I continue trying to understand all the changes, what's the status with the tests? Any estimate on whether this is a fundamental issue or just about working through the various scattered small issues here and there?

src/som/interpreter/Method.java
smarr: Not sure why I don't see these in the review, but we probably still want this to be `@CompilationFinal` for performance reasons.

src/som/interpreter/SArguments.java
```diff
@@ -58,4 +91,139 @@ public static SImmutableArray getArgumentsWithoutReceiver(final Object[] argumen
...
+    if (VmSettings.ACTOR_ASYNC_STACK_TRACE_STRUCTURE) {
+      Object[] arguments = new Object[] {receiver, null};
+      entryLoad.loadShadowStackEntry(arguments, expression, frame, false);
+      return arguments;
+    } else {
+      return new Object[] {receiver};
+    }
+  }
+
+  public static Object[] getPlain1ArgumentWithReceiver(final Object receiver,
```
smarr: In Java, we got method overloading for better or for worse. So, I don't think we need to encode arity in method names here.

src/som/interpreter/SArguments.java
```diff
@@ -58,4 +91,139 @@ public static SImmutableArray getArgumentsWithoutReceiver(final Object[] argumen
...
+    }
+  }
+
+  public static ShadowStackEntry instantiateTopShadowStackEntry(final Node expr) {
+    return ShadowStackEntry.createTop(expr);
+  }
+
+  public static ShadowStackEntry instantiateShadowStackEntry(final ShadowStackEntry previous,
+      final Node expr, final boolean async) {
+    if (async) {
```
smarr, line 187: Could you add the following? This is a compile time check that async is known and can be used for optimization. It's not a hint that this is the case, it will only result in an error if it isn't a constant value. Good for debugging of performance issues.

```suggestion
    CompilerAsserts.partialEvaluationConstant(async);
    if (async) {
```

src/som/interpreter/SArguments.java
```diff
@@ -58,4 +91,139 @@ public static SImmutableArray getArgumentsWithoutReceiver(final Object[] argumen
...
+    Object maybeShadowStack = args[args.length - 1];
+    if (maybeShadowStack instanceof ShadowStackEntry) {
+      return (ShadowStackEntry) maybeShadowStack;
+    }
+    return null;
+  }
+
+  public static void setShadowStackEntry(final Object[] args,
+      final ShadowStackEntry entry) {
+    // assert args[args.length - 1] == null : "Assume shadow stack entry is not already set.";
```
smarr: Hm, why is this commented out?
Seems like a sensible assertion, or does it not hold? If so, in which case does it fail, should the assertion be removed or documented?

src/som/interpreter/nodes/dispatch/BackCacheCallNode.java
```diff
@@ -0,0 +1,47 @@
...
+    } else if (VmSettings.ACTOR_ASYNC_STACK_TRACE_STRUCTURE) {
+      SArguments.setShadowStackEntryWithCache(arguments, expression,
+          shadowStackEntryLoad, frame, false);
+    }
+    assert arguments[arguments.length - 1] != null
+        || (frame.getArguments()[frame.getArguments().length - 1] == null)
+        || !VmSettings.ACTOR_ASYNC_STACK_TRACE_STRUCTURE;
+  }
+
+  void uniqueCaller();
```
smarr: This and the next method seem to be mutating methods. Perhaps to be clear `set` or `make` would be a good prefix for them?

src/som/interpreter/nodes/dispatch/CachedDispatchNode.java
```diff
@@ -2,19 +2,30 @@
...
 
-public final class CachedDispatchNode extends AbstractDispatchNode {
+public final class CachedDispatchNode extends AbstractDispatchNode
+    implements BackCacheCallNode {
 
-  @Child private DirectCallNode       cachedMethod;
-  @Child private AbstractDispatchNode nextInCache;
+  @Child private DirectCallNode         cachedMethod;
+  @Child private AbstractDispatchNode   nextInCache;
+  @CompilationFinal private boolean     uniqueCaller;
```
smarr: Same issue as in other node, you'll need an assumption to make this correct.

src/som/interpreter/nodes/dispatch/LexicallyBoundDispatchNode.java
```diff
@@ -1,22 +1,33 @@
...
  * Private methods are special, they are linked unconditionally to the call site.
  * Thus, we don't need to check at the dispatch whether they apply or not.
  */
-public final class LexicallyBoundDispatchNode extends AbstractDispatchNode {
+public final class LexicallyBoundDispatchNode extends AbstractDispatchNode
+    implements BackCacheCallNode {
 
-  @Child private DirectCallNode cachedMethod;
+  @Child private DirectCallNode         cachedMethod;
+  @CompilationFinal private boolean     uniqueCaller;
```
smarr: And all the other nodes: needs assumption

src/som/interpreter/Invokable.java
clementbera: Done

src/som/interpreter/SArguments.java
clementbera: So I changed to getPlainXArgumentWithReceiver with variable number of remaining parameters and made a single function instead of 3.

src/som/interpreter/SArguments.java
clementbera: It should be uncommented. Done.

src/som/interpreter/SArguments.java
clementbera: Done, it works.

src/som/interpreter/Method.java
clementbera: I tried to fix the concern with invalidation (the 4 comments) in my latest commit.

Since the field uniqueCaller is used in Method only during relinking, it is not compiled by Graal and does not have to be compilation final. The fields uniqueCaller in CachedDispatchNode and LexicallyBoundDispatchNode have however to be. I created an assumption at creation time which is invalidated when the method makeMultipleCaller (was: multipleCaller) is executed.

Looks like some of the test code I have still runs that way.

clementbera:
Before we added our change, there was already a try catch getting invalid assumption and replacing the node by the next in cache. This happened when guard.entryMatches raised the invalid assumption exception.

With new code this led to an infinite loop since the node was replaced with another node with the exact same guard.

I added a branch in the catch to check if the problem came from the guard or the stillUniqueCaller assumption. Depending on where it comes from, the problem is dealt with differently. In both cases, it leads in the end to a stable ast with a fast path.

smarr:
@clementbera ok, thanks for the explanation!

src/som/interpreter/SArguments.java
smarr: @ctrlpz please do not commit such a change. These assertions are there for a reason. It just cost me a lot of time that the assertion wasn't there.

smarr:
This PR is out of date.

A new one may be opened based on the current branch derived from Carmen's version, which is based on Clement's code: https://github.com/smarr/SOMns/tree/ct/async-stack-traces

PR #292, opened 2019-02-12
Fix overflow bug in DividePrim

fniephaus:
`Long.MIN_VALUE / -1` is the only division that overflows.
Since the result is well-known, it can be precomputed.

"In cases where the size is int or long and overflow errors need to be detected, the methods addExact, subtractExact, multiplyExact, and toIntExact throw an ArithmeticException when the results overflow. For other arithmetic operations such as divide, absolute value, increment, decrement, and negation overflow occurs only with a specific minimum or maximum value and should be checked against the minimum or maximum as appropriate."
See https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html

### DIFF
--- a/core-lib/TestSuite/IntegerTests.ns
+++ b/core-lib/TestSuite/IntegerTests.ns
@@ -195,6 +195,11 @@ class IntegerTests usingPlatform: platform testFramework: minitest = (
 
       self assert:  3 equals: 7 / 2.0.
       self assert:  2 equals: 7 / 3.5.
+
+      self assert: 9223372036854775808 equals: (* Java.MIN_VALUE *)   -9223372036854775808 / -1.
+
+      (* To trigger issue, we need to compute the value, because it is otherwise parsed as a big integer *)
+      self assert: 9223372036854775808 equals: (* Java.MIN_VALUE *) ((-922337203685477580 * 10) - 8) / -1.
     )
 
     public testDouble = (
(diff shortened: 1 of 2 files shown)


smarr:
Thanks a lot of the PR, I opened issues on the related repos.

Will have a look at it, and let the benchmarks run to get an impression of the impact of this change.
Not sure we can do much better though. And it's a correctness issue, so, yeah, difficult.

Perhaps one idea for an experiment to see how performance is impacted is to do this as a post-computation check.
Thus, check whether the result is what the overflow would give, then check the preconditions, and then deopt the thing to a specialization that produces big integers.

Todos for myself:
- [x] check benchmarks
- [ ] try post-compute guard/deopt
- [x] add a test

smarr:
@fniephaus I am confused.

I added a test, and it passes without your change.

```patch
diff --git a/core-lib/TestSuite/IntegerTests.ns b/core-lib/TestSuite/IntegerTests.ns
index 5f6aac59f..f95ba7132 100644
--- a/core-lib/TestSuite/IntegerTests.ns
+++ b/core-lib/TestSuite/IntegerTests.ns
@@ -195,6 +195,8 @@ class IntegerTests usingPlatform: platform testFramework: minitest = (
 
       self assert:  3 equals: 7 / 2.0.
       self assert:  2 equals: 7 / 3.5.
+
+      self assert: 9223372036854775808 equals: (* Java.MIN_VALUE *) -9223372036854775808 / -1.
     )
 
     public testDouble = (

```

What am I doing wrong?

smarr:
Ok, I figured it out. The first tests literal was directly parsed into BigIntegers, which hid the issue.

fniephaus:
Whoops...fun! :)

smarr (approved):

smarr:
Thanks! Benchmarks look ok. Change seems to work. So, I merged it.

PR #293, opened 2019-02-13
Snapshot replay

daumayr:
This PR implements replay from snapshots.
It also fixes various problems with the previous snapshot implementation e.g. Outer Objects of Classses, ownership of promises and various duplication issues.

TODO: enable snapshotting/serialization tests


src/som/VM.java
```diff
@@ -293,6 +293,7 @@ public void initalize(final SomLanguage lang) throws IOException {
 
     assert objectSystem == null;
     objectSystem = new ObjectSystem(new SourcecodeCompiler(lang), structuralProbe, this);
+
```
daumayr:
```suggestion
```

src/som/interpreter/actors/EventualMessage.java
```diff
@@ -79,16 +80,16 @@ public SourceSection getTargetSourceSection() {
...
-    ReceivedRootNode rm = (ReceivedRootNode) this.onReceive.getRootNode();
-
-    if (sb.needsToBeSnapshot(getMessageId())) {
-      // Not sure if this is optimized, worst case need to duplicate this for all messages
-      return rm.getSerializer().execute(this, sb);
-    } else {
-      // need to be careful, might interfere with promise serialization...
-      return -1;
+  @TruffleBoundary
+  // TODO: can we establish a structure for this? at the moment, we have an
```
smarr, line 84: This is now handled, right?

src/som/interpreter/actors/EventualMessage.java
```diff
@@ -112,7 +113,7 @@ protected AbstractDirectMessage(final Actor target, final SSymbol selector,
       this.sender = sender;
       this.target = target;
 
-      if (VmSettings.SNAPSHOTS_ENABLED) {
+      if (VmSettings.SNAPSHOTS_ENABLED && !VmSettings.REPLAY) {
```
smarr, line 116: Seeing these and other flags, should we try to have two distinct: `SNAPSHOT_RECORDING` and `SNAPSHOT_REPLAY`? This could possibly avoid issues with wrongly combining/testing flags?

So, I mean, instead of having `SNAPSHOT_ENABLED`.

src/som/vmobjects/SClass.java
```diff
@@ -191,17 +199,39 @@ public void initializeStructure(final MixinDefinition mixinDef,
     this.isTransferObject = isTransferObject;
     this.isArray = isArray;
     this.instanceClassGroup = classFactory;
+
+    if (VmSettings.SNAPSHOTS_ENABLED) {
```
smarr, line 203: Could this change go into its own method?

src/tools/snapshot/SnapshotRecord.java
```diff
@@ -45,26 +72,31 @@ public long getObjectPointer(final Object o) {
...
-
+  @TruffleBoundary
   public void addObjectEntry(final Object o, final long offset) {
     synchronized (entries) {
       entries.put(o, offset);
     }
   }
 
-  public void handleTodos(final SnapshotBuffer sb) {
+  @TruffleBoundary // TODO: convert to an approach that constructs a cache
```
smarr, line 82: Is this TODO still current?

src/tools/snapshot/deserialization/DeserializationBuffer.java
```diff
@@ -1,91 +1,222 @@
...
+
+  public boolean allreadyDeserialized(final long reference) {
+    return deserialized.containsKey(reference);
+  }
+
+  public int getNumObjects() {
+    return deserialized.size();
+  }
+
+  private void printPosition(final long current) {
```
smarr, line 75: This and the `printClass` methods look strange.
Are they used? Should they be implemented but have a VmSettings.DEBUG flag or something? I think there may be such a flag?

src/som/interpreter/actors/EventualMessage.java
daumayr: Yes, refactoring the flags probably reduces ambiguity and improves readability of the code

src/som/interpreter/actors/EventualMessage.java
daumayr: The serializer is compilation final (child node of receive).
There is still some indirection of accessing fields

src/som/vmobjects/SClass.java
daumayr: i can make an initializeForSnapshotting method

src/tools/snapshot/SnapshotRecord.java
daumayr: only if you want to use cachedserialization nodes

src/tools/snapshot/deserialization/DeserializationBuffer.java
daumayr: I haven't used them for a while, we can probably remove them.
We could also use a flag, but this can produce a LOT of output.

PR #297, opened 2019-05-28
Swap tracing buffers for blocked threads

smarr:
The forced swapping of buffers deadlocks when it encounters threads that are blocked in Java synchronization mechanisms.

This should be handled explicitly, by marking the threads, I think.

However, for the moment, we can actually check whether the thread is parked, i.e., blocked.
Unclear whether this includes all threads though.

@daumayr what do you think of this change?

### DIFF
--- a/src/tools/concurrency/TracingBackend.java
+++ b/src/tools/concurrency/TracingBackend.java
@@ -404,7 +404,9 @@ public class TracingBackend {
       this.snapshotVersion = snapshotVersion;
     }
 
-    protected TraceWorkerThread() {}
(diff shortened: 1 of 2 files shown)


daumayr:
Changes look good, was this a problem with breakpoint suspension?

smarr:
Yes, when hitting a breakpoint, we force-swap the buffers, to be able to update the visualization, and then that thread waits for the buffers to be swapped, which does not happen for threads blocked in java synchronization -> deadlock. Or well, probably a livelock actually (because the loop still goes over all threads again and again.

Took me a while to realize that.

PR #298, opened 2019-06-28
Update Truffle to >19.0.1 release

smarr:
This PR gets us onto the latest Truffle version.
There are various little changes, some of which may have bigger impact down the line.

- The interop support is now even more minimal than before, but, it seems sufficient for debugging, and for everything that's currently tested.
- This PR also restore testing on macOS.
- We use now `mx native-image` which simplifies image generation and avoids having to maintain dependencies manually.

@daumayr would be great if you could give this a spin and see whether it works on your machine.


smarr:
@fniephaus possibly relevant for you, too.

@richard-roberts @kjx @IsaacOscar @Jamigrith: this is a major update of Truffle, and we might want to consider this for https://github.com/gracelang/moth-SOMns It allows to use the officially released GraalVM 19

src/som/compiler/MixinDefinition.java
```diff
@@ -73,6 +72,7 @@
  * at runtime, which then also has the super class and mixins resolved to be
  * used to instantiate {@link SClass} objects.
  */
+@ExportLibrary(InteropLibrary.class)
```
fniephaus, line 75: Why add `@ExportLibrary(InteropLibrary.class)` to this class if you don't implement any of the interop messages (with `@ExportMessage`)?

src/som/primitives/threading/TaskThreads.java
```diff
@@ -25,6 +25,7 @@
 
 public final class TaskThreads {
 
+  @ExportLibrary(InteropLibrary.class)
```
fniephaus, line 28: Same question (from above).

src/som/vmobjects/SObjectWithClass.java
```diff
@@ -3,12 +3,13 @@
...
 import com.oracle.truffle.api.frame.MaterializedFrame;
-import com.oracle.truffle.api.interop.ForeignAccess;
+import com.oracle.truffle.api.interop.InteropLibrary;
+import com.oracle.truffle.api.library.ExportLibrary;
 
-import som.interop.SObjectInteropMessageResolutionForeign;
 import som.interpreter.objectstorage.ClassFactory;
 
 
+@ExportLibrary(InteropLibrary.class)
```
fniephaus, line 12: And again...

src/som/compiler/MixinDefinition.java
smarr: Work in progress

fniephaus: Ok, I was just wondering because this is not a draft pull request. 😉

smarr: This is not going to be addressed in this PR. There's no need for more interop support.

PR #304, opened 2019-07-29
Fix kompos tracing

daumayr:
This PR fixes the following bugs in the kompos tracing:

- duplicated Turn start/end events for actors.
- scope loss when swapping buffer. During parsing events from the replacement buffer are assigned to any open scope.
- breakpoint handing in replay used wrong message object.
- avoid constantly swapping buffers because of incorrect boolean condition

The PR also enables kompos tracing without webdebugger, and during replay executions.

### DIFF
--- a/src/som/interpreter/actors/Actor.java
+++ b/src/som/interpreter/actors/Actor.java
@@ -30,7 +30,6 @@ import tools.concurrency.TracingActors.ReplayActor;
 import tools.concurrency.TracingActors.TracingActor;
 import tools.debugger.WebDebugger;
 import tools.debugger.entities.ActivityType;
-import tools.debugger.entities.DynamicScopeType;
 import tools.replay.actors.ActorExecutionTrace;
 import tools.replay.nodes.TraceActorContextNode;
 import tools.snapshot.SnapshotBuffer;
@@ -301,15 +300,9 @@ public class Actor implements Activity {
       }
 
       try {
-        if (VmSettings.KOMPOS_TRACING) {
-          KomposTrace.scopeStart(DynamicScopeType.TURN, msg.getMessageId(),
-              msg.getTargetSourceSection());
-        }
         msg.execute();
       } finally {
-        if (VmSettings.KOMPOS_TRACING) {
-          KomposTrace.scopeEnd(DynamicScopeType.TURN);
-        }
+
       }
     }
 
--- a/src/som/interpreter/actors/Actor.java
+++ b/src/som/interpreter/actors/Actor.java
@@ -299,11 +299,7 @@ public class Actor implements Activity {
         TracingActor.handleBreakpointsAndStepping(msg, dbg, actor);
       }
 
-      try {
-        msg.execute();
-      } finally {
-
-      }
+      msg.execute();
     }
 
     private boolean getCurrentMessagesOrCompleteExecution() {
--- a/src/tools/concurrency/KomposTrace.java
+++ b/src/tools/concurrency/KomposTrace.java
@@ -5,6 +5,8 @@ import com.oracle.truffle.api.source.SourceSection;
 
 import bd.source.SourceCoordinate;
 import som.interpreter.actors.Actor;
+import som.interpreter.actors.Actor.ActorProcessingThread;
+import som.interpreter.actors.EventualMessage;
 import som.interpreter.nodes.dispatch.Dispatchable;
 import som.vm.Activity;
 import som.vm.ObjectSystem;
@@ -169,15 +171,27 @@ public class KomposTrace {
               Implementation.IMPL_CURRENT_ACTIVITY.getSize())) {
         return false;
       }
+
       super.swapStorage();
       this.lastActivity = null;
       recordCurrentActivity(current);
+
+      if (Thread.currentThread() instanceof ActorProcessingThread) {
+        EventualMessage msg = EventualMessage.getCurrentExecutingMessage();
+        ActorProcessingThread apt =
+            (ActorProcessingThread) ActorProcessingThread.currentThread();
+
+        if (apt.getCurrentActor() != null && msg != null) {
+          scopeStart(DynamicScopeType.TURN, msg.getMessageId(), msg.getTargetSourceSection());
+        }
+      }
+
       return true;
     }
 
     @TruffleBoundary
     protected boolean ensureSufficientSpace(final int requiredSpace, final Activity current) {
-      if (position + requiredSpace < VmSettings.BUFFER_SIZE) {
+      if ((position + requiredSpace) >= VmSettings.BUFFER_SIZE) {
         boolean didSwap = swapStorage(current);
         return didSwap;
       }
--- a/src/tools/concurrency/KomposTrace.java
+++ b/src/tools/concurrency/KomposTrace.java
@@ -1,5 +1,8 @@
 package tools.concurrency;
 
+import com.oracle.truffle.api.CompilerDirectives.TruffleBoundary;
+import com.oracle.truffle.api.source.SourceSection;
+
 import bd.source.SourceCoordinate;
 import som.interpreter.actors.Actor;
 import som.interpreter.nodes.dispatch.Dispatchable;
@@ -9,8 +12,6 @@ import som.vm.Symbols;
 import som.vm.VmSettings;
 import som.vmobjects.SInvokable;
 import som.vmobjects.SSymbol;
-import src.com.oracle.truffle.api.CompilerDirectives.TruffleBoundary;
-import src.com.oracle.truffle.api.source.SourceSection;
 import tools.debugger.PrimitiveCallOrigin;
 import tools.debugger.entities.ActivityType;
 import tools.debugger.entities.DynamicScopeType;
@@ -171,7 +172,6 @@ public class KomposTrace {
       super.swapStorage();
       this.lastActivity = null;
       recordCurrentActivity(current);
-
       return true;
     }
 
--- a/src/tools/concurrency/KomposTrace.java
+++ b/src/tools/concurrency/KomposTrace.java
@@ -1,8 +1,5 @@
 package tools.concurrency;
 
-import com.oracle.truffle.api.CompilerDirectives.TruffleBoundary;
-import com.oracle.truffle.api.source.SourceSection;
-
 import bd.source.SourceCoordinate;
 import som.interpreter.actors.Actor;
 import som.interpreter.actors.Actor.ActorProcessingThread;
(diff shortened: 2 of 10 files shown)


src/tools/concurrency/TracingActors.java
```diff
@@ -406,7 +406,7 @@ protected void processCurrentMessages(final ActorProcessingThread currentThread,
 
         for (EventualMessage msg : todo) {
           currentThread.currentMessage = msg;
-          handleBreakpointsAndStepping(firstMessage, dbg, a);
+          handleBreakpointsAndStepping(msg, dbg, a);
```
smarr, line 409: Ugh. This is a bad one. Thanks!

src/tools/debugger/entities/ActivityType.java
```diff
@@ -54,7 +53,7 @@ public byte getCompletionMarker() {
   }
 
   public int getCreationSize() {
-    return VmSettings.TRUFFLE_DEBUGGER_ENABLED ? (11 + TraceData.SOURCE_SECTION_SIZE) : 11;
+    return 11 + TraceData.SOURCE_SECTION_SIZE;
```
smarr, line 56: Shouldn't the condition be changed to `VmSettings.TRUFFLE_DEBUGGER_ENABLED || VmSettings.KOMPOS_TRACING`? At least that's the condition for writing the source section (see above).

src/tools/concurrency/KomposTrace.java
```diff
@@ -269,7 +283,7 @@ public void recordActivityCreation(final ActivityType entity, final long activit
       putLong(activityId);
       putShort(symbolId);
 
-      if (VmSettings.TRUFFLE_DEBUGGER_ENABLED) {
+      if (VmSettings.TRUFFLE_DEBUGGER_ENABLED || VmSettings.KOMPOS_TRACING) {
```
smarr: What's the relation ship between `KOMPOS_TRACING` and `TRUFFLE_DEBUGGER_ENABLED`?
I am not sure that they are strictly related, or are they?

You say you use `KOMPOS_TRACING` now for replay, which may or may have debugging enabled.

In the context of the code, this is likely just `KOMPOS_TRACING`, no?
(I would assume `KOMPOS_TRACING` needs to be enable to reach it.

Further related changes/questions below.

src/tools/concurrency/KomposTrace.java
```diff
@@ -293,7 +307,7 @@ private void recordEventWithIdAndSource(final byte eventMarker, final int eventS
       put(eventMarker);
       putLong(id);
 
-      if (VmSettings.TRUFFLE_DEBUGGER_ENABLED) {
+      if (VmSettings.TRUFFLE_DEBUGGER_ENABLED || VmSettings.KOMPOS_TRACING) {
```
smarr: Is `KOMPOS_TRACING` required to reach the code? same question as above.

src/tools/concurrency/TraceBuffer.java
```diff
@@ -15,7 +15,7 @@
 
   public static TraceBuffer create(final long threadId) {
     assert VmSettings.ACTOR_TRACING || VmSettings.KOMPOS_TRACING;
-    if (VmSettings.TRUFFLE_DEBUGGER_ENABLED) {
+    if (VmSettings.KOMPOS_TRACING) {
```
smarr, line 18: Is `KOMPOS_TRACING` required to reach the code? Or is `TRUFFLE_DEBUGGER_ENABLED` sufficient?

TRUFFLE_DEBUGGER_ENABLED is used when using VS code for debugging. I don't remember whether I used the trace for VS code though, I would guess not. But perhaps for activity ideas etc.

src/tools/concurrency/KomposTrace.java
```diff
@@ -169,15 +171,27 @@ boolean swapStorage(final Activity current) {
...
+        }
+      }
+
       return true;
     }
 
     @TruffleBoundary
     protected boolean ensureSufficientSpace(final int requiredSpace, final Activity current) {
-      if (position + requiredSpace < VmSettings.BUFFER_SIZE) {
+      if ((position + requiredSpace) >= VmSettings.BUFFER_SIZE) {
```
smarr, line 180: Ugh, I was constantly swapping buffers? *sigh*

src/som/interpreter/actors/Actor.java
```diff
@@ -300,17 +299,7 @@ private void execute(final EventualMessage msg,
...
 
-      try {
-        if (VmSettings.KOMPOS_TRACING) {
-          KomposTrace.scopeStart(DynamicScopeType.TURN, msg.getMessageId(),
-              msg.getTargetSourceSection());
-        }
-        msg.execute();
-      } finally {
-        if (VmSettings.KOMPOS_TRACING) {
-          KomposTrace.scopeEnd(DynamicScopeType.TURN);
```
smarr, line 311: Where does the scopeEnd go? (scopes can be nested, e.g. for transactions)

smarr (changes requested):
Great, thanks for doing these fixes!

Bye the way, was your student actively involved in creating them?
If so, it may be appropriate to credit their work ([Co-authored-by: name <name@example.com>](https://help.github.com/en/articles/creating-a-commit-with-multiple-authors)).

src/tools/concurrency/KomposTrace.java
daumayr: Yes, completely forgot about this bug

src/som/interpreter/actors/Actor.java
daumayr: ReceivedRootNode handlees scope start and end. My assumption was that you forgot to remove this code when you did some refactoring.

src/tools/concurrency/KomposTrace.java
daumayr: I see, Komos_Tracing is true when Truffle_Debugger_Enabled is true, so we can use just Kompos_Tracing.
The idea was to make the kompos tracing more standalone, i.e. can be used independent of debuggers/replay etc just like the actor tracing.

src/tools/concurrency/KomposTrace.java
daumayr: The trace doesn't containg information indicating wether sourcesections are present, we made the assumption that source section are present, hence the addition of Kompos_Tracing.

src/tools/concurrency/TraceBuffer.java
daumayr: kompos_tracing includes truffle_debugger_enabled, as kompos tracing is more indepenent now truffle_debugger_enabled is no longer sufficient.

src/tools/debugger/entities/ActivityType.java
daumayr: The source section size constant is already computed dynamically based on the VMSettings. and is zero when no source sections are present

daumayr:
I located and fixed the bugs myself, Patricks Parser showed that some things were off in the traces.

src/som/interpreter/actors/Actor.java
smarr: Ok, but then I believe the `scopeStart` you write after swapping buffers is not correct.

`scopeStart` and `scopeEnd` should be balanced. And it should be possible to order the buffers on reading so that `scopeStart` in a previous buffer isn't an issue and doesn't need to be repeated, no?

src/tools/concurrency/KomposTrace.java
smarr: Ok, I think then it should be possible to simplify all  `VmSettings.TRUFFLE_DEBUGGER_ENABLED || VmSettings.KOMPOS_TRACING` to `VmSettings.KOMPOS_TRACING`?

src/tools/debugger/entities/ActivityType.java
smarr: Right, ok, thanks

src/tools/concurrency/KomposTrace.java
daumayr: yes, i will do that

src/som/interpreter/actors/Actor.java
daumayr: The Idea was that ScopeStart implicitly ends the previous scope. If scopes can be nested we will need a different solution that keeps track of all open scopes, and leaves enough space to close the scopes at the end of the buffer

smarr: Yeah, scopes are supposed to be nest-able, at least according to [our paper](https://stefan-marr.de/downloads/dls17-marr-et-al-concurrency-agnostic-protocol-for-debugging.pdf). They are used for locks for instance, each `acquire()` opening a new scope, which means, technically, they are not even guaranteed to nest correctly. And they are used for transactions, which can be correctly nested.

daumayr: The Scope End event doesn't contain information on the scope that is ended, isn't that a problem in the case of locking?

smarr: Yeah, indeed. I suppose the current implementation assumes correct nesting then. Another bug ☹️

smarr: Ok, I guess we should assume at least support for correct nesting at the moment.

daumayr: The most simple solution would be to have a counter that indicates the number of open scopes, and use that counter when checking remaining space, inserting n endscope events on buffer swap.

smarr: Why do you need to end the scope?
And how do you distinguish buffer-swap scope closing from explicit scope ends?

daumayr: Hm, not sure about the details. I think there is some current thread info as well, we could store the open scopes when a new thread takes over, and when the original thread appears again restore the context. then we would not need close scope events and have the correct context for all events

smarr: In the current system, I do not think that scopes move threads. At least not as far as I can recall. So, it should be new buffers on the same thread only. So, ordering them should be possible, and no extra meta data needed, no?

daumayr: Yes, what i described has to be done in the parser

smarr: Ok, does the `scopeStart` then need to be removed? (cf https://github.com/smarr/SOMns/pull/304/files#r308452408)

daumayr: I think so, the parser in kompos probably needs to be checked that this is handled correctly.

smarr (approved):
Looks good to me! 👍

PR #305, opened 2019-07-31
Assisted actor debugging

daumayr:
This PR introduces a prototype for assisted debugging of actor programs.

Assisted Debugging enhances the deterministic replay of an erroneous
program by setting message breakpoints on messages that lead up to the
message that causes the error.

- added primitive markTurnErroneous
- use it for ReplayDemo
- added KomposTraceParser to be able to set breakpoints
- flags to `som` launcher to enable assisted debugging, and possibly
  limited the number of breakpoints


src/som/primitives/SystemPrims.java
```diff
@@ -216,6 +218,26 @@ public final Object doSObject(final String argument) {
     }
   }
 
+  @GenerateNodeFactory
+  @Primitive(primitive = "writeErrorMsg:")
```
smarr: This is not very descriptive and undocumented.

What exactly is this supposed to do?
Does it mark the current message as having caused an error?

I am also not sure about terminology here.
Is this the message, or is it really about the turn?

Could be named `markTurnAsErroneous`?

src/tools/parser/KomposTraceParser.java
```diff
@@ -0,0 +1,328 @@
...
+
+
+public class KomposTraceParser {
+
+  private static Set<Long> errorStrackTrace;
+
+  private static final int SOURCE_SECTION_SIZE = 8;
+
+  private enum TraceRecords {
+    ActivityCreation(11 + SOURCE_SECTION_SIZE),
```
smarr: Don't we have such constants already somewhere?

daumayr: now using constant from TraceData.java

src/som/primitives/SystemPrims.java
daumayr: renamed to markTurnErroneous and added explanatory comment

src/tools/parser/KomposTraceParser.java
```diff
@@ -0,0 +1,327 @@
...
+import tools.TraceData;
+import tools.debugger.entities.Marker;
+
+
+public class KomposTraceParser {
+
+  private static Set<Long> errorStrackTrace;
+
+  private enum TraceRecords {
+    ActivityCreation(11 + TraceData.SOURCE_SECTION_SIZE),
```
smarr: I think, these magic constants are still defined elsewhere.
Examples are:
 - https://github.com/smarr/SOMns/blob/385926d94f7dd3037e51e713316fd8b8b39b501b/src/tools/debugger/entities/ActivityType.java#L56-L57
 - https://github.com/smarr/SOMns/blob/385926d94f7dd3037e51e713316fd8b8b39b501b/src/tools/debugger/entities/PassiveEntityType.java#L36-L37

I would really prefer if these magic numbers are defined exactly once, and that we then simply refer to those where needed.
Otherwise, we are bound to run into trouble when we start changing things.

smarr (changes requested):

smarr (approved):
Thanks a lot for removing magic numbers.

smarr:
@Restoro @daumayr thanks for the work!
I squashed the various things into a single commit, and edited the commit message to be a bit more descriptive.

Will merge it.

Thanks!

PR #308, opened 2019-08-05
Multi-paradigm replay:  CSP

daumayr:
This PR prepares the existing tracing infrastructure for Multi-pardigm record and replay. Actor tracing is generalized, and new CSP tracing is implemented. To make the parsing of traces easier, a new multi pass strategy is adopted, where initially the trace is scanned for start locations of activity contexts and their ordering numbers. When an entity needs trace data, the contexts can then be parsed in the correct order without needing to sort all events.

CSP events include the Id of the channel that was used. However, this should not be necessary when each activity performs channel interactions in the same order. Assertions were added to confirm this assumption. In the future, this part of  CSP tracing could be removed to reduce event size.

The implementation allows for all kinds of activities to interact with a channel, but in practice is only for actors  and processes because other models haven't been integrated with the r&r.


.gitlab-ci.yml
```diff
@@ -34,11 +34,11 @@ replay_tests:
   script:
     - timeout 10m ant replay-tests
 
-snapshot_tests:
-  stage: full-test
-  tags: [benchmarks, infinity]
-  script:
-    - timeout 10m ant snapshot-tests
+#snapshot_tests:
```
smarr, line 38: Needs a comment, something that says why it is disabled, and when it will be enabled again (probably PR number. ideally, you could also add a todo on the related PR)

.travis.yml
```diff
@@ -67,11 +67,11 @@ install: |
 
 
 script:
-  - if [ "$TASK" = "core-tests"    ]; then $ANT core-tests    && $ANT serialization-tests && $ANT coverage; fi
+  - if [ "$TASK" = "core-tests"    ]; then $ANT core-tests    && $ANT coverage; fi #  && $ANT serialization-tests
```
smarr: Should the `serialization-tests` move to the snapshot tests?
Also would ideally have a comment.

And, you probably want to disable the whole `snapshot-tests` job earlier in the file.

build.xml
```diff
@@ -404,18 +404,19 @@
       <travis target="som-tests" />
     </target>
 
+    <!--
     <target name="serialization-tests" depends="compile">
```
smarr: Hm, I'd prefer not to comment out the target.

build.xml
```diff
@@ -451,12 +452,12 @@
       <travis target="replay-tests" />
     </target>
 
-    <target name="snapshot-tests" depends="compile">
+    <!--<target name="snapshot-tests" depends="compile">
```
smarr: Same here, simply not executing it should be sufficient, right?

build.xml
```diff
@@ -508,7 +509,7 @@
 
     <target name="coverage" depends="send-java-coverage,send-somns-coverage" />
 
-    <target name="tests" depends="core-tests,serialization-tests,replay-tests,coverage">
+    <target name="tests" depends="core-tests,replay-tests,coverage">
```
smarr, line 512: Please leave a TODO comment and the name of the target in the line above.

core-lib/Benchmarks/SavinaCSP.ns
```diff
@@ -60,4 +60,217 @@ class SavinaCSP usingPlatform: platform andHarness: harness = Value (
     public newInstance = ( ^ self new: 2000 )
     public setupVerifiedRun: run = ( run problemSize )
   )
+
+
+    public class ForkJoinThroughput new: numProcesses  totalMessages: numMessages = Benchmark <: Value (
```
smarr: Whitespace a bit off?

core-lib/TestSuite/Serialization.ns
```diff
@@ -258,5 +258,5 @@ class SerializationTests usingPlatform: platform testFramework: minitest = Value
       (assert: ppc promise resolvedWith: 42).
   )
 
-  ) : ( TEST_CONTEXT = () )
+  ) : ( DEACTIVATED_TEST_CONTEXT = () )
```
smarr: Please leave a TODO comment here to explain this.

And similar to the targets, do we really need to deactivate it here to get the desired effect?
I'd think changes to the CI setup are sufficient, no? (and perhaps in `SomTests`)

src/som/interpreter/processes/SChannel.java
```diff
@@ -70,27 +76,55 @@ public static SChannelInput create(final SynchronousQueue<Object> cell,
 
     private final SynchronousQueue<Object> cell;
     protected final SChannel               channel;
+    private final AtomicInteger            numReads;
 
     public SChannelInput(final SynchronousQueue<Object> cell,
         final SChannel channel) {
       this.cell = cell;
       this.channel = channel;
+      if (VmSettings.ACTOR_TRACING | VmSettings.REPLAY) {
```
smarr: Single or double pipe?

src/som/interpreter/processes/SChannel.java
```diff
@@ -70,27 +76,55 @@ public static SChannelInput create(final SynchronousQueue<Object> cell,
...
       try {
-        return cell.take();
+        synchronized (this) {
+          if (VmSettings.ACTOR_TRACING) {
+            traceRead.record(channel.getId(), numReads.getAndIncrement());
+          }
+          return cell.take();
+        }
       } finally {
+        if (VmSettings.REPLAY) {
```
smarr, line 102: Would be good to have a comment here why synchronized is not needed.
Or, add a synchronized, and remove the use of the atomic int.
I think, using synchronized might be more consistent, and easier to understand.
Same for the other similar code.

src/som/interpreter/processes/SChannel.java
```diff
@@ -121,26 +155,54 @@ public static SChannelOutput create(final SynchronousQueue<Object> cell,
 
     private final SynchronousQueue<Object> cell;
     protected final SChannel               channel;
+    private final AtomicInteger            numWrites;
 
     protected SChannelOutput(final SynchronousQueue<Object> cell, final SChannel channel) {
       this.cell = cell;
       this.channel = channel;
+      if (VmSettings.ACTOR_TRACING | VmSettings.REPLAY) {
```
smarr: Double `||`?

src/som/interpreter/processes/SChannel.java
```diff
@@ -70,27 +76,55 @@ public static SChannelInput create(final SynchronousQueue<Object> cell,
...
+    public Object read(final RecordTwoEvent traceRead) throws InterruptedException {
       ObjectTransitionSafepoint.INSTANCE.unregister();
+
+      if (VmSettings.REPLAY) {
+        Activity reader = TracingActivityThread.currentThread().getActivity();
+        ChannelReadRecord crr = (ChannelReadRecord) reader.getNextReplayEvent();
+        assert crr != null : reader;
+        assert channel.getId() == crr.channelId;
+
+        while (numReads.get() != crr.readNo) {
```
smarr: can you factor this out into a `replayDelay(currentId, expectedId, currentChannelId, expectedChannelId)` or so method? and use it consistently for all the places with such asserts/loops?

I hope this would make the code more readable, and avoids undesirable inconsistencies.

src/som/interpreter/processes/SChannel.java
```diff
@@ -121,26 +155,54 @@ public static SChannelOutput create(final SynchronousQueue<Object> cell,
...
+    public void write(final Object value, final RecordTwoEvent traceWrite)
+        throws InterruptedException {
       ObjectTransitionSafepoint.INSTANCE.unregister();
+
+      if (VmSettings.REPLAY) {
+        Activity reader = TracingActivityThread.currentThread().getActivity();
+        ChannelWriteRecord cwr = (ChannelWriteRecord) reader.getNextReplayEvent();
+        assert channel.getId() == cwr.channelId;
+
+        while (numWrites.get() != cwr.writeNo) {
```
smarr: Same comment as for reads.

src/som/interpreter/processes/SChannel.java
```diff
@@ -121,26 +155,54 @@ public static SChannelOutput create(final SynchronousQueue<Object> cell,
...
-        cell.put(value);
+        synchronized (this) {
+          if (VmSettings.ACTOR_TRACING) {
+            traceWrite.record(channel.getId(), numWrites.getAndIncrement());
+          }
+          cell.put(value);
+        }
+
       } finally {
+        if (VmSettings.REPLAY) {
```
smarr, line 178: Same comment as for reads.

core-lib/Benchmarks/SavinaCSP.ns
smarr: Should we add the CSP benchmarks to our usual benchmark runs/rebench setup?

I think, now that you are doing research on the CSP stuff, it would be good to make sure things don't break.

src/tools/concurrency/TracingActors.java
```diff
@@ -44,25 +44,33 @@
...
     }
 
-    protected TracingActor(final VM vm, final int id) {
+    protected TracingActor(final VM vm, final long id) {
       super(vm);
-      this.actorId = id;
+      this.activityId = id;
     }
 
     public final int getActorId() {
```
smarr, line 59: Could we just remove this method?

src/tools/replay/actors/ActorExecutionTrace.java
```diff
@@ -17,117 +18,127 @@
...
-    byte[] b = getExtDataByteBuffer(ta.getActorId(), dataId, Double.BYTES);
+    byte[] b = getExtDataByteBuffer(ta.getId(), dataId, Double.BYTES);
     TraceBuffer.UNSAFE.putDouble(
         b, TraceBuffer.BYTE_ARR_BASE_OFFSET + EXT_DATA_HEADER_SIZE, d);
     recordSystemCall(dataId, tracer);
     t.addExternalData(b);
   }
 
-  private static final int EXT_DATA_HEADER_SIZE = 3 * 4;
+  private static final int EXT_DATA_HEADER_SIZE = 16;
```
smarr: Did you add another 4byte field to the header? should it be `4 * 4` then?

smarr (changes requested):

src/tools/replay/actors/ActorExecutionTrace.java
daumayr: one of the 3 integer fields was for actor id, which is now a long. hence 4 more byte needed. I'll change it to 2*Integer.Bytes + Long.Bytes

src/tools/concurrency/TracingActors.java
daumayr: My idea was to leave it in so that I don't have to touch the snapshotting code (making rebase of snapshotting easier).

core-lib/Benchmarks/SavinaCSP.ns
daumayr: good idea, should i add a new savinaCSP suite to the codespeed.conf

src/tools/replay/actors/ActorExecutionTrace.java
smarr: Ah, yes, that would be even better!

src/tools/concurrency/TracingActors.java
smarr: Ah, right, you mentioned that. Though, could you add a comment with a todo?

core-lib/Benchmarks/SavinaCSP.ns
smarr: yes, that would be good. Thanks

src/som/interpreter/processes/SChannel.java
daumayr: added synchronization

src/som/interpreter/processes/SChannel.java
daumayr: Made it event more generic than suggested, can be reused for any numbered events in passive entities.

core-lib/Benchmarks/SavinaCSP.ns
daumayr: added codespeed conf entries and a gitlab ci job

.travis.yml
```diff
@@ -20,7 +20,7 @@ matrix:
     - env: TASK=kompos-tests
     - env: TASK=replay1-tests
     - env: TASK=replay2-tests
-    - env: TASK=snapshot-tests
+    #- env: TASK=snapshot-tests
```
smarr, line 24: A brief note on why commented out, please. Makes it easier to track this.

.travis.yml
```diff
@@ -67,11 +67,11 @@ install: |
...
 
 script:
-  - if [ "$TASK" = "core-tests"    ]; then $ANT core-tests    && $ANT serialization-tests && $ANT coverage; fi
+  - if [ "$TASK" = "core-tests"    ]; then $ANT core-tests    && $ANT coverage; fi
   - if [ "$TASK" = "checkstyle"    ]; then $ANT checkstyle    && $ANT eclipseformat-check && run cd tools/kompos && nvm install 7 && run npm install && run npm run lint; fi
   - if [ "$TASK" = "kompos-tests"  ]; then run nvm install 8  && $ANT && cd tools/kompos  && run npm -s run verify && run npm test; fi
   - if [ "$TASK" = "replay1-tests" ]; then $ANT compile       && run ./tests/replay/test.sh 1; fi
   - if [ "$TASK" = "replay2-tests" ]; then $ANT compile       && run ./tests/replay/test.sh 2; fi
-  - if [ "$TASK" = "snapshot-tests" ]; then $ANT compile      && run ./tests/snapshot/test.sh; fi
+  #- if [ "$TASK" = "snapshot-tests" ]; then $ANT compile      && run ./tests/snapshot/test.sh; && $ANT serialization-tests fi
```
smarr: I guess this doesn't strictly need to be commented out anymore.
If so, perhaps a brief note on why commented out, please. Makes it easier to track this.

codespeed.conf
```diff
@@ -689,6 +723,16 @@ experiments:
...
 
+    SOMns-SavinaCSP-exp:
+        description: Run the Savina Actor benchmarks
+        executions:
+            - SOMns-interp-tn-exp:
+                suites:
+                  - savina-csp
+            - SOMns-graal-tn-exp:
+                suites:
+                  - savina-csp
```
smarr, line 757: We might need separate configs (suites) for the Graal and interpreter version.
I try to size them so that the Graal version runs 40-80ms in the optimized case.
The interpreter version should not run too long, we don't want to wait ages. :)

smarr (changes requested):
Thanks for all the changes, think there are only a few minor things now.

.travis.yml
daumayr: uncommented it and added a comment to the other commented thing

codespeed.conf
daumayr: separated suites and scaled benchmarks. interp still uses same problem size, difference in performance is small

.travis.yml
daumayr: hm looks like this task gets executed if i uncomment this line...

smarr: looks more like a [syntax error](https://travis-ci.org/smarr/SOMns/jobs/568721883#L2019) to me:

```
/home/travis/.travis/functions: eval: line 104: syntax error near unexpected token `&&'
/home/travis/.travis/functions: eval: line 104: `if [ "$TASK" = "snapshot-tests" ]; then $ANT compile      && run ./tests/snapshot/test.sh; && $ANT serialization-tests fi '
```

I would guess the `;` before the `&&` should be removed.

codespeed.conf
smarr: Hmm, perhaps that's the benchmarks, or my naive implementation of CSP...

smarr (approved):

PR #309, opened 2019-08-06
Put Long#parseLong() behind boundary

fniephaus:
`Long#parseLong()` doesn't look like something for PE, so this puts it behind a `@TruffleBoundary`. Not tested, not benchmarked... take it or leave it. 😉

### DIFF
--- a/src/som/primitives/IntegerPrims.java
+++ b/src/som/primitives/IntegerPrims.java
@@ -76,11 +76,13 @@ public abstract class IntegerPrims {
     }
 
     @Specialization
+    @TruffleBoundary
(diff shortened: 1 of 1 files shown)


smarr:
Thanks, the PR reduced the number of methods included for runtime compilation into the native image by 16 from [2503](https://travis-ci.org/smarr/SOMns/jobs/568017159#L1316) to [2487](https://travis-ci.org/smarr/SOMns/jobs/568382083#L1331).
So, should be a good change.
Will push it to the benchmark machine shortly.

smarr (approved):

PR #312, opened 2019-08-13
Restore performance of replay

smarr:
This PR restores the performance to the same level as before #308 was merged.

The main problem introduced was that parts of the trace file were read multiple times.
This caused the replay test jobs to take 2-3 times more time on non-SSD systems.


daumayr (approved):
Thanks for fixing the parser performance, changes look good

smarr:
@daumayr thanks! Did you see the [TraceParserTests](https://github.com/smarr/SOMns/pull/312/files#diff-4a12ffad9c551776e91738dcb3a194f0)?

They are fairly minimal, but I hope this gives us a bit of confidence when stuff gets changed around.

Will merge later today or tomorrow, I hope.

daumayr:
yes, is saw the tests

PR #314, opened 2019-08-14
Send Symbols together with external data to debugger UI

ctrlpz:
Pull request that moves the call to the method `writeSymbols` after `writeExternalData` in the `TracingBackend` class when processing the trace data.  This way symbols can be sent earlier and avoid races when sending the tracing information.

### DIFF
--- a/src/tools/concurrency/TracingBackend.java
+++ b/src/tools/concurrency/TracingBackend.java
@@ -564,12 +564,13 @@ public class TracingBackend {
       while (cont || !fullBuffers.isEmpty() || !externalData.isEmpty()) {
         writeExternalData(externalDataStream);
 
+        writeSymbols(symbolStream);
+
         BufferAndLimit b = tryToObtainBuffer();
         if (b == null) {
           continue;
         }
 
-        writeSymbols(symbolStream);
         if (front != null) {
           front.sendTracingData(b.getReadingFromStartBuffer());
         }


smarr:
Thanks for the PR. I changed the based of the PR from `release` to `dev`.

ctrlpz:
yes, sorry about that, indeed the base of the PR is the `dev` branch.

smarr:
Hm, after inspecting what `tryToObtainBuffer` does, I am not sure that the change here is doing what you intent it to do.

`tryToObtainBuffer` waits to get a buffer from a queue:

https://github.com/smarr/SOMns/blob/dev/src/tools/concurrency/TracingBackend.java#L411-L419

```Java
    private BufferAndLimit tryToObtainBuffer() {
      BufferAndLimit buffer;
      try {
        buffer =
            fullBuffers.poll(VmSettings.BUFFER_TIMEOUT, TimeUnit.MILLISECONDS);
        if (buffer == null) {
          return null;
        } else {
          return buffer;
```

The time out actually blocks the thread for that many seconds if there's no buffer.
But once we got one, we send our symbols.

What your change does is to ensure that symbols are send more frequently, and possibly in smaller chunks. This is likely avoiding the race you are seeing, because the symbols arrive earlier.

However, it does not fix the actual races.
Is there a better way to avoid the race?

ctrlpz:
If this solution does not avoid completely the races then I think in the interpreter side there is not much we can do, because as we discussed before we don’t know all symbols at the beginning of the execution.

The other option I was thinking was to somehow ensure in the client side that the trace information is not resolved until I have received all the symbols, (which I think should correspond to the ones contained in the trace.ksym file). I’m not sure if in the interpreter there is a method that can provide to the client that information, e.g. send a flag to the client (together with the symbol information) to notify when all the symbols have been sent?

ctrlpz:
Looking at the code of `writeSymbols` method https://github.com/smarr/SOMns/blob/dev/src/tools/concurrency/TracingBackend.java#L527 I understand that the symbols are written in the .ksym file just after those symbols are sent to the client, then I think we don't know all the symbols at that moment neither. However, because the program execution has been paused we are not at the beginning anymore, shouldn't we know all the symbols until that point of the execution? couldn't we write the symbols first to the file and after send them?

smarr:
> If this solution does not avoid completely the races then I think in the interpreter side there is not much we can do, because as we discussed before we don’t know all symbols at the beginning of the execution.

I did not notice any issues in the Kompos UI. Though there, some operations on the trace are postponed until it is sure that entities such as actors are resolved, as far as I remember.
Could you do something similar to that?

> couldn't we write the symbols first to the file and after send them?

How would that help you? My understanding is that they are only written to the file when tracing is done, but not generally when debugging. The content of the stream and the file would be identical at that point, too. So, the race you see on the stream, is the same you'd see on the file, no?

ctrlpz:
> > If this solution does not avoid completely the races then I think in the interpreter side there is not much we can do, because as we discussed before we don’t know all symbols at the beginning of the execution.
> 
> I did not notice any issues in the Kompos UI. Though, there some operations on the trace are postponed until it is sure that entities such as actors are resolved, as far as I remember.
> Could you do something similar to that?
> 
I will check that solution in Kompos and I will try to implement it in my plugin.
> > couldn't we write the symbols first to the file and after send them?
> 
> How would that help you? My understanding is that they are only written to the file when tracing is done, but not generally when debugging. The content of the stream and the file would be identical at that point, too. So, the race you see on the stream, is the same you'd see on the file, no?

I don't think the race appears in both places, because when I got the exception, e.g. `java.lang.IndexOutOfBoundsException: Index: 1622, Size: 694` the entry for the symbol exist in the file, e.g: `1622:file:/Users/carmentorres/GitHub-repositories/IntelliJ/IntelliJ-Projects/myproject/src/programs/ShoppingApplication.ns` , however, in the list of symbols I have to registered all the symbols I received, there are only 694 elements at that moment, thus, that entry is not there yet.

smarr:
> however, in the list of symbols I have to registered all the symbols I received

Ehm, so, at the moment, we first write to the socket/stream, and then to the file.
And in both cases we write the same data.

So, I would expect any client to see the symbol first on the stream, and only afterwards in a file.

Of course, it could happen that the data from the stream is not yet processed, but it should already be there. One thing that I see for the file writing is however the `flush` operation, which should ensure that all data is written to disk. I don't know whether the same `flush` happens for the socket.

ctrlpz:
I think the flush does not happen for the socket, I have tried in [https://github.com/smarr/SOMns/blob/dev/src/tools/debugger/FrontendConnector.java#L226](url) this code: 
```
private void send(final Message msg) {
    ensureConnectionIsAvailable();
    messageSocket.send(gson.toJson(msg, OutgoingMessage.class));
    boolean flushAndClose = messageSocket.isFlushAndClose();
``` 
and it returned false. Looking here [https://github.com/TooTallNate/Java-WebSocket/blob/master/src/main/java/org/java_websocket/WebSocketImpl.java#L563](url) it seems that in the websocket implementation the **flush** is tied to the **close** operation, then it only happens if the web socket is closed as I understand. If we could call the method `onWriteDemand ` for the websocket then we could do the flush, I'm trying to see if is possible.

ctrlpz:
I have not found a way to use that method directly, however I found this related post [https://github.com/TooTallNate/Java-WebSocket/issues/779](url) I will take a look at the fragmented example, otherwise I think I will go for the promise handling solution as in Kompos.

smarr:
Hm, well, my feeling is what ever you do, you are only going to reduce the chance of a race, you are not actually resolving it.

The Kompos solution, as I understand it, actually ensures there is a synchronization edge between receive and use, as such, it solves the race, instead of reducing its likelihood.

ctrlpz:
I implemented a solution similar to the one of Kompos to avoid the races, I should test it more but at the moment it works :) I got the values from the trace without exceptions, so, I think I don't need the change of this pull request anymore, we can leave the `processTraceData` method as it is.

smarr:
Great, that's good news :)

Will close this for now.

PR #317, opened 2019-08-29
Fix bug in AsyncTestCase

ctrlpz:
Pull request to fix a bug in the `AsyncTestCase` class related to the failed assertions. The `onError` block of the `runUsing` method declares `e` as the variable for the exception but it was wrongly referenced as `ex`. Added a small test case that triggers a failed assertion in ActorTests.ns.

### DIFF
--- a/core-lib/TestSuite/ActorTests.ns
+++ b/core-lib/TestSuite/ActorTests.ns
@@ -145,6 +145,16 @@ class ActorTests usingPlatform: platform testFramework: minitest = Value (
       ^ assert: r resolvedWith: true.
     )
 
+    public testAsyncTurnResultsInNearRefFail= (
+      | bobP r |
+      bobP:: (actors createActorFromValue: Bob) <-: meet: self.
+      r:: bobP whenResolved: [:bob |
+          (bob <-: answerNearRef)
+             whenResolved: [:bob2 | bob2 = bob ]].
+      ^ assert: r resolvedWith: false.
+    )
+
+
     public testAsyncTurnResultsInValue = (
       | bob |
       bob:: (actors createActorFromValue: Bob) <-: meet: self.
--- a/core-lib/TestSuite/ActorTests.ns
+++ b/core-lib/TestSuite/ActorTests.ns
@@ -15,12 +15,6 @@ class ActorTests usingPlatform: platform testFramework: minitest = Value (
     msg:: errorTestMessage.
   )()
 
-  private class NoExceptionReceivedException signal = Exception (
-          | public msg |
-          msg:: 'received value, expected exception'.
-          self signal.
-  )()
-
   class Bob meet: a = (
     | private alice = a. |
   ) (
@@ -29,8 +23,6 @@ class ActorTests usingPlatform: platform testFramework: minitest = Value (
     public answerValue    = ( ^ 42           )
     public raiseException = ( ^ Exception signal )
   )
-
-  class Peter = () ()
 
   class DeepChain = (
     |exception|
@@ -212,22 +204,6 @@ class ActorTests usingPlatform: platform testFramework: minitest = Value (
         pp resolve: nil
       ].
       ^ pp promise
-    )
-
-    public testAsyncAssertionFailed= (
-      | bobP peterP assertP resultP whenP |
-          peterP:: (actors createActorFromValue: Peter).
-          bobP:: (actors createActorFromValue: Bob) <-: meet: peterP.
-          resultP:: bobP whenResolved: [:bob |
-                 (bob <-: answerFarRef)
-                 whenResolved: [:peter | peter = bob ]
-                 ].
-
-         assertP:: assert: resultP resolvedWith: true.
-
-         whenP:: assertP whenResolved:[:v | NoExceptionReceivedException signal ]
-                     onError: [:e | assert: e messageText equals: 'Equality assertion failed; expected: true, was: false' ].
-         ^ whenP
     )
 
   ) : (
@@ -635,6 +611,11 @@ class ActorTests usingPlatform: platform testFramework: minitest = Value (
   public class PromiseErrored = AsyncTestContext (
   | private Exception = kernel Exception. aliceValue = 10. bobValue = 20.|
   )(
+     private class NoExceptionReceivedException signal = Exception (
+        | public msg |
+        msg:: 'received value, expected exception'.
+        self signal.
+    )()
 
     private class Alice new = (
     | exception |
--- a/core-lib/TestSuite/ActorTests.ns
+++ b/core-lib/TestSuite/ActorTests.ns
@@ -15,6 +15,12 @@ class ActorTests usingPlatform: platform testFramework: minitest = Value (
     msg:: errorTestMessage.
   )()
 
+  private class NoExceptionReceivedException signal = Exception (
+          | public msg |
+          msg:: 'received value, expected exception'.
+          self signal.
+  )()
+
   class Bob meet: a = (
     | private alice = a. |
   ) (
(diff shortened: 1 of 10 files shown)


smarr:
@ctrlpz thanks for the PR.

Looks like the test you are adding is failing.
Could you change it in a way that it still tests the desired behavior but does not actually fail?

ctrlpz:
I changed the test, I reused some of the classes there was implemented in the test file. However, I think there is another bug in the tests, but in this case, related to the equality of Strings, please don't close the pull request yet, I'm looking into this other problem now.

ctrlpz:
@smarr you can finish this pull request if you agree with the test.
What I suspected that was another bug was not, I had missing in my test method name "Async" and that was the reason that comparisons of Strings in assertion declarations did not work. I found here [https://github.com/smarr/SOMns/blob/dev/core-lib/TestSuite/Minitest.ns#L415](url) the validation of the method names for using test case types `(AsyncTestCase` or `TestCase),` however,  because `AsyncTestContext` extends from `TestContext` developers can implement both async and sync test methods in their tests, e.g. [https://github.com/smarr/SOMns/blob/dev/core-lib/TestSuite/ActorTests.ns#L100](url) . I think it will be good to document somewhere this restriction of the method name starting with "testAsync" in order to be able to use correctly the assertions declared in `AsyncTestContext` .

core-lib/TestSuite/ActorTests.ns
```diff
@@ -206,6 +214,22 @@ class ActorTests usingPlatform: platform testFramework: minitest = Value (
       ^ pp promise
     )
 
+    public testAsyncAssertionFailed= (
```
smarr: Wouldn't this test fit much better into the [MinitestTests](https://github.com/smarr/SOMns/blob/dev/core-lib/TestSuite/MinitestTests.ns#L215)?
It seems misplaced here.

core-lib/TestSuite/Minitest.ns
```diff
@@ -441,12 +441,12 @@ OTHER DEALINGS IN THE SOFTWARE. *)
               exClass:: (ClassMirror reflecting: e) classObject.
               exClass == TestFailureException ifTrue: [
                 promisePair resolve: (testContextInstance
-                  ifNil:    [ TestFailure case: self description: ex messageText]
-                  ifNotNil: [:it | it createFailureResultFor: self description: ex messageText])
+                  ifNil:    [ TestFailure case: self description: e messageText ]
```
smarr: Giving that the surrounding code uses `ex` as name for the exception variable, and `exClass` for the name of the class, wouldn't it be better to stay consistent with that naming?

smarr:
> I think it will be good to document somewhere this restriction of the method name starting with "testAsync" in order to be able to use correctly the assertions declared in `AsyncTestContext` .

Yes, documentation is indeed missing.

I suppose good places would be:

- https://github.com/smarr/SOMns/blob/dev/core-lib/TestSuite/Minitest.ns#L51
- https://github.com/smarr/SOMns/blob/dev/core-lib/TestSuite/Minitest.ns#L341
- https://github.com/smarr/SOMns/blob/dev/core-lib/TestSuite/Minitest.ns#L421

ctrlpz:
@smarr please take a look at the last commit, I did some changes.
I agree with having the test in the MinitestTest file. I also simplified the test a bit, and I reused the Math class from the ActorTest file for that. I also wrote some sentences for the documentation in the locations you pointed me. Please, let me know if another change is needed.

Sometimes I used `-dnu` command to see the execution trace when I got any error, I found strange for the MinitestsTest file this output, tests did not failed however. This log corresponds to the version of the file before my changes. Not sure why if there are not failing tests triggering the trace shows a lookup problem.

```
Stack Trace
	Actors>>#λasyncdonextprevPromisecompletionResolver@173@24:    Actors.ns:177:28
	Tester>>#λrunAllAsynchronous@1180@38:                         Minitest.ns:1181:11
	Tester>>#executeAsynchronous:                                 Minitest.ns:1206:21
	AsyncTestCase>>#runUsing:                                     Minitest.ns:488:9
	Block>>#on:do:                                                Kernel.ns:649:18
	vmMirror>>#exceptionDo:catch:onException:                     ExceptionDoOnPrimFactory:1:1
	AsyncTestCase>>#λrunUsing@459@7                               Minitest.ns:482:11
	Block>>#on:do:                                                Kernel.ns:649:18
	vmMirror>>#exceptionDo:catch:onException:                     ExceptionDoOnPrimFactory:1:1
	AsyncTestCase>>#λλrunUsing@459@7@459@8                        Minitest.ns:460:21
	TestCase>>#sendTestMessage                                    Minitest.ns:415:56
	ObjectMirror>>#perform:                                       Mirrors.ns:107:45
	vmMirror>>#obj:perform:                                       PerformPrimFactory:1:1
	AssertionTests>>#testAsyncResultCreation                      MinitestTests.ns:267:16
	Tester>>#runAll                                               Minitest.ns:1186:7
	Tester>>#runAllSynchronous                                    Minitest.ns:1172:27
	Vector>>#do:                                                  Kernel.ns:688:49
	Tester>>#λrunAllSynchronous@1172@31:                          Minitest.ns:1173:33
	Tester>>#executeSynchronous:                                  Minitest.ns:1194:21
	TestCase>>#runUsing:                                          Minitest.ns:389:11
	Block>>#on:do:                                                Kernel.ns:649:18
	vmMirror>>#exceptionDo:catch:onException:                     ExceptionDoOnPrimFactory:1:1
	TestCase>>#λrunUsing@388@9                                    Minitest.ns:388:15
	TestCase>>#runUnsafeUsing:                                    Minitest.ns:427:9
	Block>>#on:do:                                                Kernel.ns:649:18
	vmMirror>>#exceptionDo:catch:onException:                     ExceptionDoOnPrimFactory:1:1
	TestCase>>#λrunUnsafeUsing@421@7                              Minitest.ns:421:8
	TestCase>>#sendTestMessage                                    Minitest.ns:415:56
	ObjectMirror>>#perform:                                       Mirrors.ns:107:45
	vmMirror>>#obj:perform:                                       PerformPrimFactory:1:1
	ResultCreationTestContext>>#testErrorCreation                 MinitestTests.ns:147:11
	TestContext>>#assert:                                         Minitest.ns:689:7
	TestContext>>#assert:description:                             Minitest.ns:693:7
	TestContext>>#assert:descriptionBlock:                        Minitest.ns:697:21
	ResultCreationTestContext>>#λtestErrorCreation@147@19         MinitestTests.ns:147:22
Lookup of #zork failed in Integer```

ctrlpz:
Hi @smarr, I have pushed some changes to the `MinitestTests` in order to run the asynchronous tests. These changes are failing because the `AssertionTests` contains mutable slots. Perhaps I’m missing something about the Tester class, but following  the current implementation for the other tests the tester instance needs the declaration of the test catalog, in order to know the test suite to test, and the corresponding slots in the `AssertionTests` class are considered as mutable, which bring to this error `Run-time Error: MinitestTests.ns:13:3: The class AssertionTests is declared as value, but also declared mutable slots`.
Please, when you have some minutes take a look at these changes and let me know if you see something wrong.

smarr:
Debugging Notes:

 - the tests fail with a `NotAValue` error, which I tracked down to the `TestCatalog`
 - The `TestCatalog` contains a `Dictionary`, which is mutable, because we do not have a version that's a `Value`
 - I fixed some inconsistencies in what's reported by `NotAValue` error

smarr:
@Marmat21 I think this could also use a look from you.
I think it's now ready for merging.

smarr:
Will merge. Review still welcome.

PR #322, opened 2019-10-09
Use `.equals()` for String comparison in guard

fniephaus:


fniephaus:
Build failures seem to be unrelated (something `npm`-related broke).

fniephaus:
SOMns relies on String interning here, so identity check should be fine.

smarr:
To provide some further rational:

SOMns is indeed supposed to use string interning for selectors.
This means, there can be bugs around not interning selector strings, and, worse, one can construct programs in a way that there are selector memory leaks.

However, the reference equality is on purpose.
Using `.equal()` makes the failure cases depend on the content of the two strings, which possibly involves a loop to compare them.
In a tight loop, and a dispatch cache with multiple entries, this would make the failure case rather slow.

I am sure there are bugs in the system around this. Though, I'd prefer those bugs to be fixed instead of changing the system's design.

PR #323, opened 2019-10-11
Add SOMns applications and tests

ctrlpz:
- add five SOMns applications and tests. These applications feature asynchronous message passing using actors. The goal is to use them later for a user study to evaluate the debugging features for actors.

### DIFF
--- a/core-lib/demos/applications/InstantMessengerApplication.ns
+++ b/core-lib/demos/applications/InstantMessengerApplication.ns
@@ -1,0 +1,165 @@
+class InstantMessengerApplication usingPlatform: platform = Value (
+| private actors = platform actors.
+  private Array = platform kernel Array.
+  private TransferArray = platform kernel TransferArray.
+  private Dictionary = platform collections Dictionary.
+  private TransferObject = platform kernel TransferObject.
+|
+)(
+
+public class InstantMessenger new: name  total: size = (
+  | private buddyMap ::= Dictionary new: size.
+    private name = name.
+    private textMessage ::= nil.
+  |
+)(
+
+    public startChat: remoteMessenger = (
+      | pDiscover pName pSend msg pp array |
+
+        pp:: actors createPromisePair.
+
+        pDiscover:: self addMessenger: remoteMessenger. (* returns a far reference of the remote messenger *)
+
+        pName:: pDiscover <-: getName.
+        pName whenResolved: [:remoteName |
+                msg:: ('Hello ' + remoteName).
+                pSend:: sendMessage: remoteName contentMsg: msg.
+
+                pSend whenResolved:[:result |
+                    (result = 'ok')
+                    ifTrue:[
+                         array:: TransferArray new: 2. (* array with the msg to display in the sender actor screen *)
+                         array at: 1 put: name.
+                         array at: 2 put: msg.
+                         pp resolver resolve: array.
+                    ]
+                    ifFalse:[
+                         pp resolver resolve: nil.
+                    ]
+                ].
+        ].
+
+
+        ^ pp promise
+    )
+
+    public getName = (
+      ^ name
+    )
+
+    public sendMessage: receiverName contentMsg: content = (
+      | pp receiverActor pReceive|
+
+      textMessage:: TextMessage new: content sender: getName.
+
+      receiverActor:: buddyMap at: receiverName.
+      pReceive:: receiverActor <-: receive: textMessage.
+
+      ^ pReceive
+    )
+
+    public displayMessage: msg = (
+      (getName +' screen> '+ msg) println
+    )
+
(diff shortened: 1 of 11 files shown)


core-lib/demos/applications/InstantMessengerApplication.ns
```diff
@@ -0,0 +1,165 @@
...
+
+      (* print message to the screen *)
+      self displayMessage:  sender +': '+ content.
+
+      (* acknowledgement notification of message received *)
+      ^ 'ok'.
+    )
+)
+
+ class TextMessage new: content sender: senderName = TransferObject (
```
smarr: Hm, is there a specific reason why this is a TransferObject?
Seems like the slots are immutable, so, it looks like this could be a `Value`?

core-lib/demos/applications/InstantMessengerApplication.ns
```diff
@@ -0,0 +1,165 @@
...
+  | private msg = content.
```
smarr: Since these slots are immutable, it seems unnecessary to have getters (`getContent`, `getSender`). Instead, these slots could simply be made `public`.

core-lib/demos/applications/InstantMessengerApplication.ns
```diff
@@ -0,0 +1,165 @@
...
+
+        pName:: pDiscover <-: getName.
+        pName whenResolved: [:remoteName |
+                msg:: ('Hello ' + remoteName).
+                pSend:: sendMessage: remoteName contentMsg: msg.
+
+                pSend whenResolved:[:result |
+                    (result = 'ok')
+                    ifTrue:[
+                         array:: TransferArray new: 2. (* array with the msg to display in the sender actor screen *)
```
smarr: You got a `TextMessage` class here, which seems to fulfill the same purpose. Would it make sense to use it instead?

core-lib/demos/applications/InstantMessengerApplication.ns
```diff
@@ -0,0 +1,165 @@
...
+                         pp resolver resolve: nil.
+                    ]
+                ].
+        ].
+
+
+        ^ pp promise
+    )
+
+    public getName = (
```
smarr: `name` is immutable. The getter seems unnecessary.

core-lib/demos/applications/InstantMessengerApplication.ns
```diff
@@ -0,0 +1,165 @@
...
+
+        pDiscover:: self addMessenger: remoteMessenger. (* returns a far reference of the remote messenger *)
+
+        pName:: pDiscover <-: getName.
+        pName whenResolved: [:remoteName |
+                msg:: ('Hello ' + remoteName).
+                pSend:: sendMessage: remoteName contentMsg: msg.
+
+                pSend whenResolved:[:result |
+                    (result = 'ok')
```
smarr: Instead of using strings, in Smalltalk, it would be more idiomatic to use a Symbol: `#ok`.

core-lib/demos/applications/InstantMessengerApplication.ns
```diff
@@ -0,0 +1,165 @@
...
+      | sender content p |
+
+      content:: textMessage getContent.
+      sender:: textMessage getSender.
+
+      (* print message to the screen *)
+      self displayMessage:  sender +': '+ content.
+
+      (* acknowledgement notification of message received *)
+      ^ 'ok'.
```
smarr: The earlier comment about the Symbol, would apply here as well.

smarr (commented):
Just looking at the first app

core-lib/demos/applications/InstantMessengerApplication.ns
smarr: Looking at the tests, this seems indeed be a value object. So, I changed it.

core-lib/demos/applications/InstantMessengerApplication.ns
smarr: fixed

core-lib/demos/applications/InstantMessengerApplication.ns
smarr: Fixed

core-lib/demos/applications/InstantMessengerApplication.ns
smarr: Fixed.

core-lib/demos/applications/InstantMessengerApplication.ns
smarr: Fixed.

core-lib/demos/applications/InstantMessengerApplication.ns
smarr: Fixed.

PR #331, opened 2019-11-28
Fix for STM bugs

daumayr:
- Issue 1:
The BlockDispatchNode used in Primitives like `to: do: []` did not use atomic call targets, breaking the transactional semantic. Fixed by specializing BlockDispatchNode and using the appropriate calltargets.

- Issue 2:
Up to now the commit lock was only held during the conflict check part of the commit protocol. Application of changes to shared state was performed without global synchronization.
As a result two conflicting transaction of which one should fail may both succeed to commit.
This is possible because the application of changes by the first transaction races with the consistency check by the second transaction.

- Issue 3:
Conflict detection for primitive slots caused problems in the vacation benchmark. Check of Object slots also checked primitive slots with identity comparison. Fixed by filtering primitive slots as they are already checked separately before the object slots.

- Added STM versions of the Vacation and Lee benchmarks

STM overhead in Lee is ~5x mutex. STM is slightly faster than mutex for Vacation.

### DIFF
--- a/core-lib/Benchmarks/LeeSTM.ns
+++ b/core-lib/Benchmarks/LeeSTM.ns
@@ -1,10 +1,11 @@
 (* *)
-class LeeTM usingPlatform: platform andHarness: harness = (
+class LeeSTM usingPlatform: platform andHarness: harness = (
 | private Benchmark = harness Benchmark.
   private Array     = platform kernel Array.
   private Vector    = platform kernel Vector.
   private Mutex     = platform threading Mutex.
   private Thread    = platform threading Thread.
+  private Transaction = platform transactions Transaction.
   private system    = platform system.
 
   private EMPTY      = 0.
@@ -196,7 +197,6 @@ class LeeTM usingPlatform: platform andHarness: harness = (
   class LeeRouter new: data gridSize: size = (
   | public gridSize  = size.
     private grid     = Grid width: size height: size depth: 2.
-    private gridLock = Mutex new.
 
     private netNo      ::= 0.
     private failures   ::= 0.
@@ -206,7 +206,6 @@ class LeeTM usingPlatform: platform andHarness: harness = (
     private inputLine ::= nil.
     private linePos   ::= 1.
 
-    private queueLock = Mutex new.
     private work      = WorkQueue new.
   |
     parseData: data.
@@ -252,7 +251,7 @@ class LeeTM usingPlatform: platform andHarness: harness = (
     )
 
     public nextTrack = (
-      queueLock critical: [
+      Transaction atomic: [
         work next notNil ifTrue: [
           ^ work dequeue
       ] ].
@@ -263,7 +262,7 @@ class LeeTM usingPlatform: platform andHarness: harness = (
       | done |
       (* start transaction *)
       done:: false.
-      gridLock critical: [
+      Transaction atomic: [
         done:: connectXs: q x1 ys: q y1 xg: q x2 yg: q y2 netNo: q nn tempg: tempg ].
       (* end transaction *)
       ^ done.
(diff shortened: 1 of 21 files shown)


smarr:
Could you try before and after with the LeeTM benchmark?

There's a version with transactions here: https://github.com/MetaConc/SOMns/tree/lee-tm

daumayr:
Yes, thanks for pointing out what to try it with.

smarr:
I had a very brief first look.

I got only one request for consistency:

`LeeTM` seems misnamed, it doesn't use transactions, does it?
Could you rename it to simply `Lee`?

For the STM versions, could you drop the `_` from the file names to be consistent with the names we show in codespeed? Just to avoid confusion.

daumayr:
ok, I'll change that.

smarr:
@daumayr I pushed a test case last night. and ran everything in Kent.

Unfortunately, I got an issue, haven't investigated further yet. Seems a little strange:

```
Executing run:LeeSTM (SOMns-interp, macro-startup, 1 0 4) 1 None None
    cmd: /home/gitlab-runner/builds/d258e35c/0/sm951/SOMns/som -G -t1   core-lib/Benchmarks/Harness.ns LeeSTM  1 0 4
    cwd: /home/gitlab-runner/builds/d258e35c/0/sm951/SOMns
    Run failed. Return code: 2
    Output:

        Uncaught exception on ActorProcessingThread-0
        Processing failed for: Actor
        Stack Trace: (printing may fail in some situation with a null pointer exception
        java.lang.NullPointerException
        	at som.compiler.MethodBuilder.determineOuterMixinIds(MethodBuilder.java:706)
        	at som.compiler.MethodBuilder.getOuterRead(MethodBuilder.java:688)
        	at som.compiler.Parser.outerSend(Parser.java:1278)
        	at som.compiler.Parser.primary(Parser.java:1223)
        	at som.compiler.Parser.binaryOperand(Parser.java:1389)
        	at som.compiler.Parser.formula(Parser.java:1466)
        	at som.compiler.Parser.keywordMessage(Parser.java:1427)
        	at som.compiler.Parser.messages(Parser.java:1332)
        	at som.compiler.Parser.evaluation(Parser.java:1140)
        	at som.compiler.Parser.setterSends(Parser.java:1119)
        	at som.compiler.Parser.expression(Parser.java:1097)
        	at som.compiler.Parser.blockBody(Parser.java:1072)
        	at som.compiler.Parser.blockContents(Parser.java:987)
        	at som.compiler.Parser.methodBlock(Parser.java:912)
        	at som.compiler.Parser.methodDeclaration(Parser.java:849)
        	at som.compiler.Parser.sideDeclaration(Parser.java:728)
        	at som.compiler.Parser.classBody(Parser.java:509)
```

smarr:
@daumayr I think https://github.com/smarr/SOMns/pull/331/commits/f0c109e70851e9558dbb8ef196441d7018aec68e fixes the issue mentioned earlier.

Running CI now.

core-lib/Benchmarks/LeeSTM.ns
```diff
@@ -0,0 +1,603 @@
...
+      data isNil ifTrue: [ loadData: problemSize ].
+
+      totalLaidTracks:: 0.
+      lr:: LeeRouter new: data gridSize: gridSize.
+
+      threads:: Array new: numThreads withAll: [
+        Thread spawn: [ | t |
+          t:: LeeThread new: lr.
+          t run.
+          m critical: [
```
smarr: @daumayr hm, looks like you're still using a mutex here. Is this on purpose?

daumayr: no , i thought i used a search function to get rid of all of them. apparently i missed one.

daumayr: fixed

smarr (approved):

PR #332, opened 2019-12-19
Run Unit tests on Substrate VM build

### LINKED ISSUE #330: Run Unit tests on Substrate VM build
Would be good to run the [basic SOMns tests](https://github.com/smarr/SOMns/blob/dev/build.xml#L416-L419) on the created native executable. We should probably add an extra target `som-native-tests` for that.

However, before we can do that, we should make sure that the `som-native` binary is built in the root directory, possibly by specifying a path here: https://github.com/smarr/SOMns/blob/dev/build.xml#L547

And, we also need to detect whether we run on `som-native` to disable the unsupported extension tests.

For that, it would be good to have a new primitive, possibly `vmMirror featuresSupportsExtensions: nil`.
It should probably be made available as `System>>#supportsExtensions` here: https://github.com/smarr/SOMns/blob/dev/core-lib/System.ns

An implementation should return `true` when executing on Hotspot, and `false` when executing on Substrate.

The [ExtensionTests](https://github.com/smarr/SOMns/blob/dev/core-lib/TestSuite/ExtensionTests.ns) then somehow need to check for that. Though, I don't think we have a mechanism for it yet. Possibly for the moment just testing the flag and returning from the test if extensions are not supported.

sophie-kaleba:
Fix #330 - allows the som-native executable to run the basic SOMNs tests.

- native executable is built on the root dir of the project
- new target `native-tests` to run the tests on the native executable
- new primitive `supportsExtensions` that checks the VM in use. `ExtensionsTests` are skipped if SubstrateVM is used


smarr (approved):
Thanks for this!

I added a few bits, and changed how we know whether extensions are support (mostly to simplify).

Most importantly, I actually enabled the tests on Travis CI and GitLab CI.
I also changed SupportExtensionPrim to use `TruffleOptions.AOT`, which does not require a substitution and says more explicitly what happens.

Please have a look at the changes, and let me know whether it's find to merge it.

Thanks!

sophie-kaleba:
The changes seem fine to me, thanks!

smarr:
And merged. Thanks!

PR #341, opened 2020-01-28
Specify path to VM settings to black-diamonds

sophie-kaleba:
Otherwise VM settings passed on the command line are ignored when building a native image


smarr:
Thanks for fixing this!

PR #343, opened 2020-02-18
Do not cache final layout assumptions

fniephaus:
There's no need to cache a final value, so this should reduce cache sizes a bit.

The only downside is that the better name (`isLatestLayout`) is dropped, maybe `getAssumption()` should be renamed at some point (e.g. `getIsLatestLayoutAssumption()`)?

### DIFF
--- a/src/som/interpreter/objectstorage/InitializerFieldWrite.java
+++ b/src/som/interpreter/objectstorage/InitializerFieldWrite.java
@@ -89,7 +89,7 @@ public abstract class InitializerFieldWrite extends ExprWithTagsNode {
   }
 
   @Specialization(
-      assumptions = {"isLatestLayout"},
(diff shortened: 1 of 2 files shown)


smarr:
Thanks.

I guess this would require careful benchmarking?

It also introduces an extra indirection+method call in the interpreter, right?

fniephaus:
I think this should produce the same code when running with Graal. In terms of interpretation speed, you may be right. But when running without the compiler, you wouldn't want assumptions anyway, so I think it's ok to drop these caches. Also, we'll only know if this reduces memory consumption if you measure it as part of your benchmarking ;)

smarr:
Results are too flaky to know what's going on. Though, I'll merge it anyway.

fniephaus:
Performance shouldn't change really, only memory usage should slightly be reduced (at least in theory).

PR #344, opened 2020-02-28
Replace BranchProfile with transferToInterpreterAndInvalidate

fniephaus:
The `ArithmeticException` triggers a rewrite, so it is ok to just transferToInterpreter.

### DIFF
--- a/src/som/primitives/IntegerPrims.java
+++ b/src/som/primitives/IntegerPrims.java
@@ -2,11 +2,11 @@ package som.primitives;
 
 import java.math.BigInteger;
 
+import com.oracle.truffle.api.CompilerDirectives;
 import com.oracle.truffle.api.CompilerDirectives.TruffleBoundary;
 import com.oracle.truffle.api.dsl.GenerateNodeFactory;
 import com.oracle.truffle.api.dsl.Specialization;
 import com.oracle.truffle.api.instrumentation.Tag;
-import com.oracle.truffle.api.profiles.BranchProfile;
 
 import bd.primitives.Primitive;
 import som.interpreter.nodes.nary.BinaryComplexOperation;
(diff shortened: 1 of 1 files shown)


src/som/primitives/IntegerPrims.java
```diff
@@ -89,14 +89,13 @@ public final Object doSClass(final SSymbol argument) {
...
   public abstract static class LeftShiftPrim extends ArithmeticPrim {
-    private final BranchProfile overflow = BranchProfile.create();
 
     @Specialization(rewriteOn = ArithmeticException.class)
     public final long doLong(final long receiver, final long right) {
       assert right >= 0; // currently not defined for negative values of right
 
       if (Long.SIZE - Long.numberOfLeadingZeros(receiver) + right > Long.SIZE - 1) {
-        overflow.enter();
+        CompilerDirectives.transferToInterpreter();
```
smarr: Does this need to be this?:

```suggestion
        CompilerDirectives.transferToInterpreterAndInvalidate();
```

My understanding is that the code only properly gets invalidated when it's done when going from the compiled code to the interpreter. If I remember correctly, @chumer may have mentioned that a long time ago.

fniephaus: Invalidation follows in the generated code anyway (specialization is replaced in executeAndSpecialize), so it doesn't really matter which one you want to use.

chumer: Yes it does. It i safe to use transferToInterpreterAndInvalidate here.

smarr: @chumer: sorry, your reply is ambiguous to me.

Is `transferToInterpreter` sufficient? I would assume `transferToInterpreterAndInvalidate` definitely is, but if I recall correctly, a couple of years back, there was an issue with first doing `transferToInterpreter` and then some invalidations done in the interpreter would not trigger correctly.

chumer: > Is transferToInterpreter sufficient?

It would be wrong here. Invalidation would not be happening, as after transferToInterpreter all Truffle directives no longer work.

You should basically always `transferToInterpreterAndInvalidate` as this allows us to flow guards in the IR. There are *very* few use-cases for transferToInterpreter. To the extend that you can probably just ignore it and get rid of uses of it entirely. Even for assertions you should use transferToInterpreterAndInvalidate as it allows us to move the guard in the IR. In places where slow-paths need to be cut use BranchProfile + TruffleBoundary.

I have an open issue to improve the docs on these APIs. Even, possibly, remove one of the two.

smarr: Thanks for confirming, that's what I thought.

fniephaus: > after transferToInterpreter all Truffle directives no longer work.

Now that you say it, it makes sense...removing all uses of `transferToInterpreter` in GraalSqueak 🏃‍♂️💨

PR #345, opened 2020-04-07
Fix problems with blocks in variable declaration expressions 

smarr:
This fixes an issue with using blocks in variable declarations:

```
public testSelfInBlock = (
      | var = [ self ] value. |
      self assert: var is: self.
)
```

Before the PR, the block caused an exception, caused by how we manage variables on the MethodScope objects.

We expect them to be set exactly once, when the variable declaration parsing is completed.

However, if we have a block, the block needs a marker for us to know whether it's on the stack, which is an `Internal` variable added to the variable array.
The array may have been set (if we have a block in the method after the variable declarations), or it may not have been set yet (if we have a block in the variable declarations).
This fixes the error when we haven't set the variables yet.

@daumayr @sophie-kaleba could one of you review this and approve the PR, or suggest improvements, perhaps missing tests?


smarr:
If you have comments, or suggestions for improvements, I'll create a new PR.

Thanks

PR #348, opened 2020-04-13
Add support for actor-language features to Dynamic Metrics tool

smarr:
The goal of this PR is to improve the Dynamic Metric tool to be able to measure all actor-related language features.

Specifically, this means to measure:

- [x] actor creation, and the polymorphism of the value from which actors are created
- [x] message sends (to far refs and promises)
- [x] polymorphism of the activated methods (num different receiver types, num different target methods, activation counts) from a message send (same metrics as for normal method dispatch)
- [x] whether promises are captured, i.e., used, or simply ignored
- [x] callbacks registered on promises, type of callback operation (whenResolved, onError, whenResolvedOnError)
- [x] callback activations
- [x] explicitly created promise pairs
- [x] number of implicit promise resolutions (return from async message)
- [x] number of explicit promise resolutions (Resolver>>#resolve:)
- [x] number of promises created (implicit/explicit)
- [x] turns executed in the benchmark

For later:
- [ ] promise chain length
- [ ] messages sent per turn
- [ ] number of turns per actor
- [ ] messages received, messages sent per actor

In addition to the main changes, the PR does some maintenance:
- add double support for AbsPrim  https://github.com/smarr/SOMns/pull/348/commits/75a3df3bbdae4b32dd437663cbbcfd75d47f7e37
- avoid initialize TraceActorCreationNode if tracing is not enabled https://github.com/smarr/SOMns/pull/348/commits/f21d95966db37cb876e4653f423d473384905c42
- instead of relying on instrumentation, use simply a counting node to count direct call target activations
- fix breakpoint buttons not appearing in Kompos https://github.com/smarr/SOMns/pull/348/commits/352e0e6f9559ab281487d0b2afe24389c596611b
- make object safepoint handling safer and more consistent https://github.com/smarr/SOMns/pull/348/commits/dd07d8787e04bda48e4748f5ea18d5d039a03bf2

This is not all that could be done. More work is tracked with #349.


smarr:
@daumayr I am merging this, but have a look at for instance https://github.com/smarr/SOMns/pull/348/commits/dd07d8787e04bda48e4748f5ea18d5d039a03bf2 and https://github.com/smarr/SOMns/commit/f21d95966db37cb876e4653f423d473384905c42 which may have an impact on you.

PR #373, opened 2020-09-02
Fix Nondeterministic Corner Case in Receiver-Side Replay

daumayr:
This pull request fixes a corner case in which the receiver-side rr is not able to capture and reproduce message orderings accurately.

The problem can occur when the following conditions are met:
- a message send (M1) to a resolved promise (P1) and a promise resolution (P2) race with each other
- the resolver of P1 is performing the resolution of P2
- the actor sending M1 also sent a message M2 that is stored in P2

Under the old replay, both messages would have the same sender/receiver promise message tuple in the trace. Hence we do not capture wether M1 or M2 is processed first.

The new solution sequentially numbers all promise messages sent by an actor, and records that number in place of the id of the resolver. replay executions automatically generate the same message numbers. In combination with the sender, this allows us to uniquely identify the promise messages and accurately reorder them.

Promise resolver tracking was removed from tracing promises. 
The messageId field from Kompos tracing was reused to store message numbers.

TODO: add a testcase, do thorough testing

### DIFF
--- a/src/som/interpreter/actors/ReceivedRootNode.java
+++ b/src/som/interpreter/actors/ReceivedRootNode.java
@@ -13,7 +13,6 @@ import som.interpreter.SArguments;
 import som.interpreter.SomLanguage;
 import som.interpreter.actors.EventualMessage.PromiseMessage;
 import som.interpreter.actors.SPromise.SResolver;
-import som.interpreter.actors.SPromise.STracingPromise;
 import som.vm.VmSettings;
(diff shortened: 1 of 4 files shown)


smarr:
@daumayr do you happen to have other code related to this?

daumayr:
Sorry, I don't have code for this case, it was just something that i realized during writing. I can make you a sequence diagram if you want

PR #380, opened 2021-04-21
Fix the profiling of block activations in Dynamic Metrics

smarr:
Since we changed the block implementation to use the TruffleDSL, our stats on closure activations where incomplete.
We were missing the number of targets and activation counts.

This PR fixes the issue by collecting the statistics using a manual traversal of the DSL node structure to extract the counts from the CountingDirectCallNodes.

This is brittle and depends on the TruffleDSL not changing too much, but I suppose, there are no major changes expected.

@sophie-kaleba please review and check whether this solves your issue.


smarr:
Sigh. This was already solved in https://github.com/smarr/SOMns/pull/351 and in a better, less brittle way. Past me was clearly more alert.

